PR #10, opened 2025-09-28
Assignment 1

PigPig-SDK:
Worked with @KorbinWeiler

Extra credit: Program now allows the user up to 2 test retakes. Each retake randomizes the questions (both by question order, and answer order).

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -92,7 +92,8 @@ public class ProgramTests
     [DataRow(-500)]
     public void RandomizeOrder_ResponseMaintainsContinuity(int seed)
     {
-        //Collect a range of possible answers
+        // Test randomization with different seeds to ensure the correct answer remains in the expected position 
+        // across multiple iterations
         for (int i = 0; i < 1000; i++)
         {
             // Arrange
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -1,3 +1,5 @@
+using static PrincessBrideTrivia.Program;
+
 namespace PrincessBrideTrivia.Tests;
 
 [TestClass]
@@ -70,18 +72,43 @@ public class ProgramTests
     }
 
     [TestMethod]
-    [DataRow(10, "30%", 1, 2, 3)]
-    [DataRow(10, "80%", 3, 8, 5)]
-    [DataRow(10, "0%", 0, 0, 0)]
-    [DataRow(10, "70%", 7, 7, 5)]
-    [DataRow(10, "90%", 9, 0, 0)]
-    public void FindHighestScore_ReturnsExpectedScore(int numOfQuestions, string expectedScore, params int[] scores)
+    [DataRow("1", ResponseValidation.Accept)]
+    [DataRow("2", ResponseValidation.Reject)]
+    [DataRow("foo", ResponseValidation.Invalid)]
+    [DataRow("11", ResponseValidation.Invalid)]
+    [DataRow("1 1", ResponseValidation.Invalid)]
+    public void AcceptRetryQuiz_ReturnsExpectedResponse(string testUserInput, ResponseValidation desiredOutput)
     {
-        string highestScore = Program.FindHighestScore(scores, numOfQuestions);
-        Assert.AreEqual(expectedScore, highestScore);
+        // Arrange
+
+        // Act
+
+        // Assert
+        Assert.AreEqual(AcceptRetryQuiz(testUserInput), desiredOutput);
     }
-
-
+    [TestMethod]
+    [DataRow(69)]
+    [DataRow(5924502)]
+    [DataRow(-500)]
+    public void RandomizeOrder_ResponseMaintainsContinuity(int seed)
+    {
+        //Collect a range of possible answers
+        for (int i = 0; i < 1000; i++)
+        {
+            // Arrange
+            Question question = new Question();
+            question.CorrectAnswerIndex = "1";
+            string expectedCorrectAnswer = "1";
+            question.Answers = new string[] { "1", "2", "3" };
+            // Act
+            question.RandomizeOrder(seed + i);
+            // Assert
+            if(!string.Equals(expectedCorrectAnswer, question.Answers[int.Parse(question.CorrectAnswerIndex) - 1]))
+            {
+                Assert.Fail();
+            }
+        }
+    }
     private static void GenerateQuestionsFile(string filePath, int numberOfQuestions)
     {
         for (int i = 0; i < numberOfQuestions; i++)
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -69,6 +69,18 @@ public class ProgramTests
         Assert.AreEqual(expectedString, percentage);
     }
 
+    [TestMethod]
+    [DataRow(10, "30%", 1, 2, 3)]
+    [DataRow(10, "80%", 3, 8, 5)]
+    [DataRow(10, "0%", 0, 0, 0)]
+    [DataRow(10, "70%", 7, 7, 5)]
+    [DataRow(10, "90%", 9, 0, 0)]
+    public void FindHighestScore_ReturnsExpectedScore(int numOfQuestions, string expectedScore, params int[] scores)
+    {
+        string highestScore = Program.FindHighestScore(scores, numOfQuestions);
+        Assert.AreEqual(expectedScore, highestScore);
+    }
+
 
     private static void GenerateQuestionsFile(string filePath, int numberOfQuestions)
     {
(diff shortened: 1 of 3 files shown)


c-stanton:
I will review this

c-stanton (commented):
Nice! Everything works.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -1,27 +1,79 @@
...
+        Console.WriteLine($"Your final score is {GetPercentCorrect(attemptScores.Max(), questions.Length)}");
     }
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
     {
-        return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+        if (numberOfQuestions == 0)
+            return "N/A";
+
+        double precent = (double)numberCorrectAnswers / numberOfQuestions * 100;
```
katmilton, line 75: Just pointing out a small typo here and on the following line. It should be 'percent'.

katmilton (commented):
## Assignment Details

* Pull Request title includes "Assignment 1" in the title ✔
* Partner has also made a commit ✔
* Pull Request targets Assignment 1 branch ✔
* Issue 1: Application no longer crashes ✔
* Issue 2: Unit test properly passes ✔

## Extra Credit

* Come up with a new feature for the application
  * Issue 3: Add a new feature request implemented ✔
  * Issue 3: Feature request unit tested ✔
  
 Your code looks nice. I really had to scrape the bottom of the barrel to find any issues. Your added features are pretty nifty!

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -69,7 +71,45 @@ public void GetPercentCorrect_ReturnsExpectedPercentage(int numberOfCorrectGuess
         Assert.AreEqual(expectedString, percentage);
     }
 
+    [TestMethod]
+    [DataRow("1", ResponseValidation.Accept)]
+    [DataRow("2", ResponseValidation.Reject)]
+    [DataRow("foo", ResponseValidation.Invalid)]
+    [DataRow("11", ResponseValidation.Invalid)]
+    [DataRow("1 1", ResponseValidation.Invalid)]
+    public void AcceptRetryQuiz_ReturnsExpectedResponse(string testUserInput, ResponseValidation desiredOutput)
```
Joshua-Lester3, line 80: This won't count against you, especially because the test names given in the original branch don't necessarily follow this either (I put a PR out to fix it): For this, and other test method names, follow the naming convention of `MethodUnderTest_ConditionUnderTest_ExpectedResult`.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -69,7 +71,45 @@ public void GetPercentCorrect_ReturnsExpectedPercentage(int numberOfCorrectGuess
...
+        // Act
 
+        // Assert
+        Assert.AreEqual(AcceptRetryQuiz(testUserInput), desiredOutput);
+    }
+    [TestMethod]
+    [DataRow(69)]
+    [DataRow(5924502)]
+    [DataRow(-500)]
+    public void RandomizeOrder_ResponseMaintainsContinuity(int seed)
```
Joshua-Lester3, line 93: This test would probably go better in its own file, since it's testing a file separate from `Program.cs` (ex: `QuestionTests.cs`).

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -69,7 +71,45 @@ public void GetPercentCorrect_ReturnsExpectedPercentage(int numberOfCorrectGuess
...
+    {
+        // Test randomization with different seeds to ensure the correct answer remains in the expected position 
+        // across multiple iterations
+        for (int i = 0; i < 1000; i++)
```
Joshua-Lester3, line 97: I definitely like the thoroughness of the testing here! :) Though, I would say a for loop to test isn't necessary in this case. One concern would be it would iterate 1000 times for each DataRow. It probably wouldn't slow the testing down by much (under the PR's checks tab, the tests only take 6s total), but probably would be bad practice going forward. I think I recall Mark saying in my class a couple years ago that one generally shouldn't have logic in tests (if statements, for loops, etc.) - though I'm sure there are exceptions.

Another concern I have: I would say each iteration is basically doing the same thing. The beginning state before randomizing is that the question.Answers are in a certain order and we assume the question.CorrectAnswerIndex is actually correct (as set by the already implemented logic). After randomizing, the code could check that the new order of question.Answers corresponds to the new question.CorrectAnswerIndex. Any testing of the same process after that would be redundant: if the code has already ensured question.CorrectAnswerIndex is correct, then at the start of the next loop would be back at the beginning state. So, that'd be doing the same test 1000 times over. I'd say just doing it once would be sufficient! :)

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -1,27 +1,78 @@
-﻿namespace PrincessBrideTrivia;
+﻿using System.Diagnostics;
+using System.Linq;
```
Joshua-Lester3, line 2: nit: Unused using statements.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -1,27 +1,78 @@
...
+
+namespace PrincessBrideTrivia;
 
 public class Program
 {
+    public enum ResponseValidation
```
Joshua-Lester3, line 8: I think enums are generally better kept in their own file. Not sure we've gone over them, or the convention to generally keep separate types in separate files. Though an argument could definitely be made that since this is only used in the Program.cs file, that it's fine here :)

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -1,27 +1,78 @@
...
         Question[] questions = LoadQuestions(filePath);
-
         int numberCorrect = 0;
-        for (int i = 0; i < questions.Length; i++)
+        int[] attemptScores = new int[maxAttempts];
+        for (int attemptIndex = 0; attemptIndex < maxAttempts; attemptIndex++)
         {
-            bool result = AskQuestion(questions[i]);
-            if (result)
+            foreach (Question question in questions)
```
Joshua-Lester3, line 31: Love the use of foreach here! I don't recall if we've gone over that, but if we haven't, good use of unknown syntax!

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -34,9 +85,31 @@ public static bool AskQuestion(Question question)
...
+        }
+        else if (userInput == "2")
+        {
+            return ResponseValidation.Reject;
+        }
+        else
+        {
+            return ResponseValidation.Invalid;
+        }
+    }
```
Joshua-Lester3, line 112: Although we haven't gone over it yet (I'm pretty sure), a switch statement would be good here!

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -1,27 +1,78 @@
...
+    public enum ResponseValidation
+    {
+        Accept,
+        Reject,
+        Invalid
+    }
+
+    private const int QuestionBlockSize = 5;
+
+    private const int AnswerOptionsGiven = QuestionBlockSize - 2;
```
Joshua-Lester3, line 17: Good job refactoring! Though, I'll note that `const` is better for fields that never change (ex: pi or feetInAMile). I could see these being changed if a later Trivia.txt was introduced.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -1,27 +1,78 @@
...
+                Console.WriteLine("1 : Yes");
+                Console.WriteLine("2 : No");
+                ResponseValidation responseValidation = ResponseValidation.Invalid;
+                while (responseValidation == ResponseValidation.Invalid) 
+                { 
+                    responseValidation = AcceptRetryQuiz(Console.ReadLine());
+                }
+                if (responseValidation == ResponseValidation.Reject)
+                {
+                    break;
```
Joshua-Lester3, line 54: This is possibly a code smell. I might refactor here. Possibly using an overarching while loop instead of a for loop.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -51,10 +124,10 @@ public static bool DisplayResult(string userGuess, Question question)
 
     public static void DisplayQuestion(Question question)
     {
-        Console.WriteLine("Question: " + question.Text);
+        Console.WriteLine($"Question: {question.Text}");
```
Joshua-Lester3, line 127: Good swap to string interpolation here and below. That'd possibly be a good change to make a PR for that Ben and Mark might accept into their Assignment1 branch.

Joshua-Lester3 (commented):
## Assignment Details

* Pull Request title includes "Assignment 1" in the title ✔
* Partner has also made a commit ✔
* Pull Request targets Assignment 1 branch ✔
* Issue 1: Application no longer crashes ✔
* Issue 2: Unit test properly passes ✔

## Extra Credit

* Come up with a new feature for the application
  * Issue 3: Add a new feature request implemented ✔
  * Issue 3: Feature request unit tested ✔

Joshua-Lester3:
@PigPig-SDK @KorbinWeiler Also, a small note for the future, be sure to resolve the peer review comments you get (fellow classmates and Copilot, not mine/other graders as that's after the Final Corrections deadline). I see they've been addressed, so I'll do no deductions, but future PRs I will. Also, do the same for the Copilot comments.

PR #22, opened 2025-10-01
Assignment 1

RyanHirte:
Worked on with @kellanburns

Program compiles and unit tests pass.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -86,6 +86,7 @@ public class Program
             question.Answers[1] = answer2;
             question.Answers[2] = answer3;
             question.CorrectAnswerIndex = correctAnswerIndex;
+            questions[i] = question;
         }
         return questions;
     }
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -21,7 +21,7 @@ public class Program
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
     {
-        return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+        return ((double)numberCorrectAnswers / numberOfQuestions * 100) + "%";
     }
 
     public static bool AskQuestion(Question question)
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -21,7 +21,8 @@ public class Program
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
     {
-        return ((double)numberCorrectAnswers / numberOfQuestions * 100) + "%";
+        double percentCorrect = ((double)numberCorrectAnswers / numberOfQuestions * 100);
+        return ((int)percentCorrect + "%");    
     }
 
     public static bool AskQuestion(Question question)
@@ -38,6 +39,11 @@ public class Program
         do
         {
             input = Console.ReadLine();
+            if (input == null)
+            {
+                Console.WriteLine("Null input received. Exiting.");
+                Environment.Exit(1); // suggested by AI, but seems reasonable here
+            }
             if (int.TryParse(input, out int guess) && guess >= 1 && guess <= 3)
             {
                 return input;
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -34,7 +34,16 @@ public class Program
 
     public static string GetGuessFromUser()
     {
-        return Console.ReadLine();
+        string input;
+        do
+        {
+            input = Console.ReadLine();
+            if (int.TryParse(input, out int guess) && guess >= 1 && guess <= 3)
+            {
+                return input;
+            }
+            Console.WriteLine("Please enter a valid option (1, 2, or 3).");
+        } while (true);
     }
 
     public static bool DisplayResult(string userGuess, Question question)


ulises-aguilar (commented):
Pull Request title includes "Assignment 1" in the title ✔
Partner has also made a commit ✔
Pull Request targets Assignment 1 branch ✔
Issue 1: Application no longer crashes ✔
Issue 2: Unit test properly passes ✔

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -21,7 +21,7 @@ public static void Main(string[] args)
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
```
ulises-aguilar, line 22: I recommend adding a check in case the numberofQuestions = 0. I'd do something like this 
if (numberOfQuestions == 0)
        {
            return "N/A";
        }

JosephPotapenko:
---

### Right now output will look like 66.66667%. You can format to one decimal place for readability:

```suggestion
double percent = (double)numberCorrectAnswers / numberOfQuestions * 100;
return $"{percent:F1}%";

**Comment by JosephPotapenko** [comment]
### Instead of string concatenation with `+`, you can use interpolation for cleaner and more readable output:

```suggestion
Console.WriteLine($"You got {GetPercentCorrect(numberCorrect, questions.Length)} correct");

**Comment by JosephPotapenko** [comment]
### `CorrectAnswerIndex` is currently a string. Using an int can avoid possible issues and make validation easier:

```suggestion
question.CorrectAnswerIndex = int.Parse(lines[lineIndex + 4]);

**Comment by JosephPotapenko** [comment]
### Right now, invalid input like `"Inigo Montoya"` is treated as incorrect. You could instead, validate that the input is numeric and within range which could avoid mistyped answers. This makes the grading more accurate and the program less prone to accidents/mistakes:

```suggestion
public static string GetGuessFromUser()
{
    string? input;
    do
    {
        input = Console.ReadLine();
        if (int.TryParse(input, out int guess) && guess >= 1 && guess <= 3)
        {
            return input;
        }
        Console.WriteLine("Please enter a valid option (1, 2, or 3).");
    } while (true);
}
```
### Though not completely necessary, this could make the program more user friendly and the results less likely to be inaccurate.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
RyanHirte: Hey Ulises, would you be able to clarify why this would be necessary? I could be wrong but I don't think it is. The questions are being generated from a file, and unless you simply don't want the program to work as intended, there has to be at least one question in the file created by the programmer, not user input. So as long as there is a proper file containing a question, this will never divide by 0 right? Thanks!

RyanHirte:
> ### Right now, invalid input like `"Inigo Montoya"` is treated as incorrect. You could instead, validate that the input is numeric and within range which could avoid mistyped answers. This makes the grading more accurate and the program less prone to accidents/mistakes:
> ```
> public static string GetGuessFromUser()
> {
>     string? input;
>     do
>     {
>         input = Console.ReadLine();
>         if (int.TryParse(input, out int guess) && guess >= 1 && guess <= 3)
>         {
>             return input;
>         }
>         Console.WriteLine("Please enter a valid option (1, 2, or 3).");
>     } while (true);
> }
> ```
> 
> ### Though not completely necessary, this could make the program more user friendly and the results less likely to be inaccurate.

Hey Joseph this is a good point and I will add this thanks! I do think it's necessary because the user can input the correct answer by typing it out and it won't be considered correct.

RyanHirte:
> ### Instead of string concatenation with `+`, you can use interpolation for cleaner and more readable output:
> ```
> Console.WriteLine($"You got {GetPercentCorrect(numberCorrect, questions.Length)} correct");
> ```

Not sure if this is necessary because this doesn't change the output.

RyanHirte:
> ### `CorrectAnswerIndex` is currently a string. Using an int can avoid possible issues and make validation easier:
> ```
> question.CorrectAnswerIndex = int.Parse(lines[lineIndex + 4]);
> ```

In the Question class (Question.cs), the correct answer index is supposed to be a string:
 `public string CorrectAnswerIndex { get; set; }` 
So correct me if I'm wrong but I don't think this is necessary.

RyanHirte:
> ### Right now output will look like 66.66667%. You can format to one decimal place for readability:
> ```
> double percent = (double)numberCorrectAnswers / numberOfQuestions * 100;
> return $"{percent:F1}%";
> ```

I think this could be a matter of preference and not make the program better or worse so I don't think this is necessary either.

RyanHirte:
@kellanburns Can you weigh in on these too? Maybe you have a different opinion?

kellanburns:
I've been following the thread, I agree with your analyses. Most of the changes appear to be matters of preference, or future-proofing edge cases that aren't present in the given unit tests. All tests pass, and the program works as intended. The change that you made for the input validation seems *valid* and with the original intent of the program.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
BraydenMcMahon: I could be wrong but I think that the program should be able to be versatile  and not rely on having a correct .txt file. If this program wants to be used for a different quiz and that file is incorrectly inputted then the program will crash when it divided by zero so maybe either catch exception is better than a crash in this case.

BraydenMcMahon (commented):
Pull Request title includes "Assignment 1" in the title ✔
Partner has also made a commit ✔
Pull Request targets Assignment 1 branch ✔
Issue 1: Application no longer crashes ✔
Issue 2: Unit test properly passes ✔

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
ulises-aguilar: > Hey Ulises, would you be able to clarify why this would be necessary? I could be wrong but I don't think it is. The questions are being generated from a file, and unless you simply don't want the program to work as intended, there has to be at least one question in the file created by the programmer, not user input. So as long as there is a proper file containing a question, this will never divide by 0 right? Thanks!

I was thinking the same thing BraydenMcMahon wrote. Although you're correct for this program, it's not really necessary.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -34,7 +35,21 @@ public static bool AskQuestion(Question question)
...
-        return Console.ReadLine();
+        string input;
+        do
+        {
+            input = Console.ReadLine();
+            if (input == null)
+            {
+                Console.WriteLine("Null input received. Exiting.");
+                Environment.Exit(1); // suggested by AI, but seems reasonable here
+            }
```
Joshua-Lester3, line 46: I don't think this will work how you want it to. If input is null, then the whole application will exit. I'd recommend testing out what Copilot says - it's very often wrong. You could check for null, but retry grabbing input if it was.

Joshua-Lester3 (commented):
## Assignment Details

* Pull Request title includes "Assignment 1" in the title ✔
* Partner has also made a commit ✔
* Pull Request targets Assignment 1 branch ✔
* Issue 1: Application no longer crashes ✔❌ I'd say this one isn't completely ruled out. I'm not sure the conditions under which Console.ReadLine might return null, but if it somehow did, the application would crash.
* Issue 2: Unit test properly passes ✔

## Extra Credit

* Come up with a new feature for the application
  * Issue 3: Add a new feature request implemented ❌ Not implemented
  * Issue 3: Feature request unit tested ❌ Not implemented

Joshua-Lester3:
@kellanburns @RyanHirte Remember to resolve and reply to all peer review comments - Copilot's review as well! :)

RyanHirte:
@Joshua-Lester3 Will do that right now sorry.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
RyanHirte: Ah I see what you guys mean, but I'm not sure it's necessary for this since it's on the programmer to provide a working file. Thank you for the review.

Joshua-Lester3:
> @Joshua-Lester3 Will do that right now sorry.

No problem! I think it's too late for this assignment - but remember it for next one :)

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -21,7 +21,7 @@ public static void Main(string[] args)
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
     {
-        return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+        return ((double)numberCorrectAnswers / numberOfQuestions * 100) + "%";
```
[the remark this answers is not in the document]
RyanHirte: This was not necessary so did not add.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -34,7 +34,16 @@ public static bool AskQuestion(Question question)
 
     public static string GetGuessFromUser()
     {
-        return Console.ReadLine();
+        string input;
+        do
+        {
+            input = Console.ReadLine();
```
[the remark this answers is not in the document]
RyanHirte: This was added because we thought it was necessary

RyanHirte:
@Joshua-Lester3 Ah okay just finished doing it, thanks for the heads up and will make sure to remember for assignment 2!

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
kellanburns: I tried playing around with it when I added it (and included the note to show my source), and I kept it because the behavior was not explicitly defined in the assignment instructions and I think that exiting on `null` input seems logical here.

Joshua-Lester3: @kellanburns I like that you're thinking about logical-ness (if that's a word lol) here. I think I'll disagree since you already have input validation below on lines 47-50, and I think this should fall into that. If anything, I think throwing an exception describing the circumstances, instead of a Console.WriteLine() would be better. But I think end user input should be validated and given feedback for if wrong and retried. I'd avoid throwing exceptions if possible because of end user input.

Anyway, thanks for replying! This kind of discussion makes both of us think about it more, and get to a better solution.

JosephPotapenko:
Didn't submit this part originally:

Assignment Details
Pull Request title includes "Assignment 1" in the title ✔
Partner has also made a commit ✔
Pull Request targets Assignment 1 branch ✔
Issue 1: Application no longer crashes ✔❌what Joshua said, even though I wouldn't have noticed that if I was completely honest.
Issue 2: Unit test properly passes ✔

PR #41, opened 2025-10-12
Assignment 3

chanderlud:
I worked with @PigPig-SDK on this assignment. We completed both extra credit options.


CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
Hellothereyoko, line 1: You declared treat warnings as errors twice. This may mess up your config, but I'd double check just to be safe...

CanHazFunny/CanHazFunny/JokeService.cs
Hellothereyoko, line 1: Not sure why you're using a Regex call. That's a system level command that shouldn't be used in a simple program like this. I think this & JokeService is where you're losing easy points.

CanHazFunny/CanHazFunny/Jester.cs
Hellothereyoko, line 1: Your approach to filtering Chuck Norris is simple. It can't account for instances where his first name is used for the joke. I would use contains and separate into cases that he could appear in a joke. I know its tedious and more code; but it'll cover your bases...

CanHazFunny/CanHazFunny/Program.cs
Hellothereyoko, line 1: Why is main sealed? Not wrong, just curious as to the reasoning...

global.json
Hellothereyoko, line 1: Good catch! This part took me a couple of days to find. Nicely done!

Hellothereyoko (commented):
I see the potential for this program to score WAY higher. Look at assignment reqs carefully & look to your classmate's implementation for inspiration. Good Hustle!  :)

CanHazFunny/CanHazFunny/JokeService.cs
chanderlud: Certain jokes contain escaped characters; this regex call unescapes those characters so that the joke prints out to the console nicely. I think I saw only one joke like this when I was testing, though.

CanHazFunny/CanHazFunny/Program.cs
chanderlud: I thought that Visual Studio had suggested this change, but when I went back and removed sealed, I did not get any new messages or warnings. I'll go ahead and remove that.

CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
chanderlud: The options `TreatWarningsAsErrors` and `CodeAnalysisTreatWarningsAsErrors` are different, I think, the first option is for the compiler while the second is for the IDE.

CanHazFunny/CanHazFunny/Jester.cs
chanderlud: I guess we never saw any jokes that contained only part of Chuck Norris' name. I would also see a potential case where the joke uses the word "chuck" as a verb and would get blocked if the filtering used each part of his name separately.

CanHazFunny/CanHazFunny/Program.cs
chanderlud: Ah right, yeah doesn't pass github actions checks without the sealed attribute. The error code is CA1852

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,175 @@
...
+    public void Constructor_WithNullService_ThrowsArgumentNullException()
+    {
+        // Arrange
+        Mock<IJokeOutput> outputMock = new();
+
+        // Act
+        ArgumentNullException ex = Assert.Throws<ArgumentNullException>(() => new Jester(outputMock.Object, null!));
+
+        // Assert
+        Assert.Equal("jokeService", ex.ParamName);
```
KorbinWeiler: Id argue that these Asserts are not necessary as if the exception is thrown It should be obvious which parameter caused the exception.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,175 @@
...
+    {
+        // Arrange
+        Mock<IJokeOutput> outputMock = new();
+        Mock<IJokeService> serviceMock = new();
+        int callCount = 0;
+
+        serviceMock.Setup(s => s.GetJoke()).Returns(() =>
+        {
+            callCount++;
+            if (callCount > 5)
```
KorbinWeiler: This seems like a really ugly way to do this. I think an alternative method that would improve this is to create a function that converts the text to a default case so that you can check what the output is, then you can use that new function in your TellJoke function to filter the case insensitivities.

CanHazFunny/CanHazFunny/IJokeOutput.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace CanHazFunny;
+
+/// <summary>
+/// Defines a contract for outputting jokes to a destination,
+/// such as the console or a text writer.
+/// </summary>
+public interface IJokeOutput
+{
+    /// <summary>
+    /// Writes the specified joke to the output destination.
```
steeley21: Great work on these interfaces! Small nit pick from the coding guidelines to avoid comments unless they describe something that is not obvious to someone other than the developer who wrote the code. Not sure these comments are necessary here and with your IJokeService interface.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,175 @@
...
+        // Assert
+        serviceMock.Verify(s => s.GetJoke(), Times.Exactly(3));
+        outputMock.Verify(o => o.PrintJoke("A clean, safe, corporate-approved joke."), Times.Once);
+    }
+
+    [Theory]
+    [InlineData("Chuck Norris once roundhouse kicked a server into uptime.")]
+    [InlineData("I heard chuck norris debugs in production.")]
+    [InlineData("CHUCK NORRIS DOES NOT NEED UNIT TESTS.")]
+    public void TellJoke_FiltersCaseInsensitively(string badJoke)
```
KorbinWeiler: This test function does not follow the standard of "method under test" "Conditions of test" "Expected results"

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,175 @@
...
+
+            return badJoke;
+        });
+
+        Jester jester = new(outputMock.Object, serviceMock.Object);
+
+        // Act + Assert
+        InvalidOperationException ex = Assert.Throws<InvalidOperationException>(() => jester.TellJoke());
+        Assert.Equal("infinite chuck norris loop", ex.Message);
+        outputMock.Verify(o => o.PrintJoke(It.IsAny<string>()), Times.Never);
```
KorbinWeiler: I dont think this is necessary since the only time this will not be true is when serviceMock does not iterate 5 times, in which case the InvalidoperationsException will not get thrown and the test will fail before it even reaches this line.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
KorbinWeiler: Magic number 5 is confusing. Instead maybe have a bool called "HasBeenRepeated" then that you change to true on an iteration and then if the value is true, then the InvalidOperationException is thrown.

CanHazFunny/CanHazFunny/IJokeOutput.cs
```diff
@@ -0,0 +1,14 @@
...
+    /// </summary>
+    /// <param name="joke">The joke text to display.</param>
+    void PrintJoke(string joke);
```
steeley21: Similar note that you could abstract this to just print() or write() to decouple from joke only outputs.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,21 @@
...
+    public IJokeService JokeService { get; } = jokeService ?? throw new ArgumentNullException(nameof(jokeService));
+    public string BannedSubstring { get; set; } = "chuck norris";
+
+    public void TellJoke()
+    {
+        string filteredJoke;
+        do
+        {
+            filteredJoke = JokeService.GetJoke();
+        } while (filteredJoke.Contains(BannedSubstring, StringComparison.OrdinalIgnoreCase));
```
steeley21, line 17: Consider adding a max attempts check here to avoid an infinite loop if the joke service continues to return chuck norris jokes.

steeley21 (commented):
For this assignment do the following:

    Create two interfaces. One to represent the JokeService and one interface to represent outputting the joke to the screen (Console.WriteLine). ✔
    The JokeService will need to have the interface applied to it. ✔
    Create an implementation of the output interface that writes out the joke to the console. ✔
    Implement the Jester class. It should take in both interfaces as dependencies. These dependencies should be null checked. ✔
    The Jester class TellJoke() method should retrieve a joke from the JokeService. If the joke contains "Chuck Norris", skip it and get another. The joke should be written to the output dependency. ✔
    Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔

Fundamentals

    Be sure you enable:
        Nullability for all projects ✔
        Ensure that you turn on treat warnings as errors for all projects ✔
        Ensure that you turn on CodeAnalysisTreatWarningsAsErrors ✔
        Ensure that you turn on EnforceCodeStyleInBuild ✔
    Ensure there are no errors or warnings (including code analysis warnings) ✔
    All of the above should be unit tested 

Extra Credit

    Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)?✔
    The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string.✔

Great work guys!!

CanHazFunny/CanHazFunny/JokeOutput.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.IO;
+
+namespace CanHazFunny;
+
+
+public class JokeOutput(TextWriter? writer = default) : IJokeOutput
+{
+    public TextWriter Writer { get; } = writer ?? Console.Out;
```
KorbinWeiler: I think this kind of output class that has multiple output types would be kind of confusing for an outside user to use. I think it would be better to change this to either a JokeFileOutput and null check the TextWriter or change it to a JokeConsoleOutput and take no input. This will make the intention of the class a lot more clear.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace CanHazFunny;
+
+public class Jester(IJokeOutput jokeOutput, IJokeService jokeService)
+{
+    public IJokeOutput JokeOutput { get; } = jokeOutput ?? throw new ArgumentNullException(nameof(jokeOutput));
+    public IJokeService JokeService { get; } = jokeService ?? throw new ArgumentNullException(nameof(jokeService));
```
KorbinWeiler: This works but I believe ArgumentNullException.ThrowIfNull() is the preffered method of throwing on a null value after .net 7.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,21 @@
...
+    public string BannedSubstring { get; set; } = "chuck norris";
```
KorbinWeiler, line 9: Feel free to ignore but it might be a good idea to make this a list of strings incase more banned substrings are added in the future

CanHazFunny/CanHazFunny/IJokeOutput.cs
KorbinWeiler: PrintJoke should probably be renamed to OutputJoke as not every form of output with be easily translated to Print.

KorbinWeiler (commented):
Create two interfaces. One to represent the JokeService and one interface to represent outputting the joke to the screen (Console.WriteLine). ✔
The JokeService will need to have the interface applied to it. ✔
Create an implementation of the output interface that writes out the joke to the console. ✔❌ (JokeOutput both outputs to the console and doesn't so I am leaving this with both a check and an x)
Implement the Jester class. It should take in both interfaces as dependencies. These dependencies should be null checked. ✔
The Jester class TellJoke() method should retrieve a joke from the JokeService. If the joke contains "Chuck Norris", skip it and get another. The joke should be written to the output dependency. ✔
Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔
Fundamentals
Be sure you enable:
Nullability for all projects ✔
Ensure that you turn on treat warnings as errors for all projects ✔
Ensure that you turn on CodeAnalysisTreatWarningsAsErrors ✔
Ensure that you turn on EnforceCodeStyleInBuild ✔
Ensure there are no errors or warnings (including code analysis warnings) ✔
Extra Credit
Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)?✔
The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string.✔

CanHazFunny/CanHazFunny/IJokeOutput.cs
PigPig-SDK: Adjusted. Good point

PigPig-SDK: Adjusted.

CanHazFunny/CanHazFunny/IJokeOutput.cs
PigPig-SDK: Agreed. Thank you for the feedback.

JosephPotapenko:
Create two interfaces. One to represent the JokeService and one interface to represent outputting the joke to the screen (Console.WriteLine). ✔  IJokeService and IOutput.
The JokeService will need to have the interface applied to it. ✔  JokeService implements IJokeService.
Create an implementation of the output interface that writes out the joke to the console. ✔ implemented and outputs to Console.Out
Implement the Jester class. It should take in both interfaces as dependencies. These dependencies should be null checked. ✔ Jester takes in IOutput and IJokeService and null checks.
The Jester class TellJoke() method should retrieve a joke from the JokeService. If the joke contains "Chuck Norris", skip it and get another. The joke should be written to the output dependency. ✔
Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔
Fundamentals
Be sure you enable:
Nullability for all projects ✔  <Nullable>enable</Nullable> included in both .csproj files.
Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔ net9.0 and LangVersion set to latest
Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ✔ AnalysisLevel is latest
Ensure that you turn on treat warnings as errors for all projects ✔ <TreatWarningsAsErrors>true</TreatWarningsAsErrors> set in both projects
Ensure that you turn on CodeAnalysisTreatWarningsAsErrors ✔
Ensure that you turn on EnforceCodeStyleInBuild ✔ present in both .csproj files
Ensure there are no errors or warnings (including code analysis warnings) ✔  no remaining build or style warnings
All of the above should be unit tested ✔ JesterTests and JokeOutputTests validate proper behaviors as far as I can tell
Choose simplicity over complexity ✔ pretty clean and straightforward

Extra Credit
Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)? ✔ tests to make sure the console outputs as far as I can tell
The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string. - didn't go through all the logic, but  HttpClient.GetFromJsonAsync<JokeResponse> used, so it looks like it was attempted.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
PigPig-SDK: Fair. Adjusted.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
PigPig-SDK: Solved.

PigPig-SDK: Used bool to simplify. I think this solution is fine enough.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
PigPig-SDK: Renamed. Thank you.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
PigPig-SDK: Sure.

CanHazFunny/CanHazFunny/JokeOutput.cs
PigPig-SDK: Agree, in technicality it might not satisfy the assignment. 
Moved the textwriter to testing.

CanHazFunny/CanHazFunny/Jester.cs
PigPig-SDK: Going to ignore for now, as I'm trying to strictly satisfy the assignments requirements (Keeping it simple).

CanHazFunny/CanHazFunny/Jester.cs
PigPig-SDK: Not sure if thats allowed during assignment..? Because the ThrowIfNull() returns void.  Going to ignore for now.

CanHazFunny/CanHazFunny/Jester.cs
PigPig-SDK: Though this could be a good idea. I'd rather keep it simple, as the assignment tells me to choose simplicity over complexity. If the user wanted jokes without chuck norris, i'd suggest they supply a better IJokeService.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,166 @@
...
+        // Arrange
+        Mock<IOutput> outputMock = new();
+        Mock<IJokeService> serviceMock = new();
+
+        // Act
+        Jester jester = new(outputMock.Object, serviceMock.Object);
+
+        // Assert
+        Assert.Same(outputMock.Object, jester.JokeOutput);
+        Assert.Same(serviceMock.Object, jester.JokeService);
```
Joshua-Lester3, line 23: Love to see some mocking :)

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,166 @@
...
+    {
+        // Arrange
+        Mock<IOutput> outputMock = new();
+
+        // Act / Assert
+        ArgumentNullException ex = Assert.Throws<ArgumentNullException>(() => new Jester(outputMock.Object, null!));
+    }
+
+    [Fact]
+    public void TellJoke_CallsGetJoke_And_PrintJoke_WithReturnedValue()
```
Joshua-Lester3, line 50: Follow the `MethodUnderTest_ConditionOfTest_ExpectedResult` test name convention - only 2 underscores.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,166 @@
...
+        serviceMock.Setup(s => s.GetJoke()).Returns(expectedJoke);
+
+        Jester jester = new(outputMock.Object, serviceMock.Object);
+
+        // Act
+        jester.TellJoke();
+
+        // Assert
+        serviceMock.Verify(s => s.GetJoke(), Times.Once);
+        outputMock.Verify(o => o.Write(expectedJoke), Times.Once);
```
Joshua-Lester3, line 66: I would also verify that jester tells the joke. Possibly through some mock magic, grabbing what the output mock is supposed to print.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,166 @@
...
+        {
+            isCalled = true;
+            if (isCalled) throw new InvalidOperationException("infinite chuck norris loop");
+            return badJoke;
+        });
+
+        Jester jester = new(outputMock.Object, serviceMock.Object);
+
+        // Act + Assert
+        InvalidOperationException ex = Assert.Throws<InvalidOperationException>(() => jester.TellJoke());
```
Joshua-Lester3, line 164: I'm not sure this is testing your implementation.

CanHazFunny/CanHazFunny.Tests/JokeOutputTests.cs
```diff
@@ -0,0 +1,39 @@
...
+        // Arrange
+        StringWriter writer = new();
+        Console.SetOut(writer);
+        JokeOutput output = new();
+
+        // Act
+        output.Write("Knock knock");
+
+        // Assert
+        Assert.Equal("Knock knock" + Environment.NewLine, writer.ToString());
```
Joshua-Lester3, line 21: Make sure you reset Console Out back to normal. I think it can affect other tests.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace CanHazFunny;
+
+public class Jester(IOutput jokeOutput, IJokeService jokeService)
+{
+    public IOutput JokeOutput { get; } = jokeOutput ?? throw new ArgumentNullException(nameof(jokeOutput));
+    public IJokeService JokeService { get; } = jokeService ?? throw new ArgumentNullException(nameof(jokeService));
+    public string BannedSubstring { get; set; } = "chuck norris";
```
Joshua-Lester3, line 9: I would make BannedSubstring readonly since it doesn't look like it would change ever.
```suggestion
    public string BannedSubstring { get; } = "chuck norris";
```

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -1,14 +1,22 @@
...
     {
-        string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api").Result;
-        return joke;
+        JokeResponse? response = HttpClient.GetFromJsonAsync<JokeResponse>("https://geek-jokes.sameerkumar.website/api?format=json").Result;
+
+        if (response is not null)
+        {
+            return Regex.Unescape(response.Joke);
+        } else {
+            throw new HttpRequestException();
```
Joshua-Lester3, line 19: I might supply a message here, to explain what went wrong.

Joshua-Lester3 (commented):
## Assignment

For this assignment do the following:

- Create *two* interfaces. One to represent the `JokeService` and one interface to represent outputting the joke to the screen (`Console.WriteLine`). ✔
- The `JokeService` will need to have the interface applied to it. ✔
- Create an implementation of the output interface that writes out the joke to the console. ✔
- Implement the `Jester` class. It should take in both interfaces as dependencies. These dependencies should be null checked. ✔
- The `Jester` class `TellJoke()` method should retrieve a joke from the `JokeService`. If the joke contains "Chuck Norris", skip it and get another. The joke should be written to the output dependency. ✔
- Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔

## Fundamentals

- Be sure you enable:
  - Nullability for all projects  ✔
  - Set `LangVersion` and the `TargetFramework` to the latest released versions available (preview versions optional) ✔
  - Ensure that you turn on code analysis for all projects(`EnableNETAnalyzers`) ✔
  - Ensure that you turn on treat warnings as errors for all projects ✔
  - Ensure that you turn on `CodeAnalysisTreatWarningsAsErrors` ✔
  - Ensure that you turn on `EnforceCodeStyleInBuild` ✔
- **Ensure there are no errors or warnings (including code analysis warnings)** ✔
- **All of the above should be unit tested** ✔
- **Choose simplicity over complexity** ✔
  
## Extra Credit

- Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)? ✔
- The Geek jokes API that is being used can also return jokes in a JSON format. Update the `JokeService` to retrieve jokes using JSON. `GetJoke` should still return a string. ✔

Overall, looks good. Nice use of Moq. You could possibly abstract out the network call (through an interface/implementation dependency so you could mock it) to make JokeService more testable, but I think the necessity of that could be argued.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
OliviaJSmith: to elaborate yeah typically one assert per test so then your name would fit the guidelines better

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,166 @@
...
+        // Act
+        jester.TellJoke();
+
+        // Assert
+        serviceMock.Verify(s => s.GetJoke(), Times.Exactly(3));
+        outputMock.Verify(o => o.Write("A clean, safe, corporate-approved joke."), Times.Once);
+    }
+
+    [Theory]
+    [InlineData("Chuck Norris once roundhouse kicked a server into uptime.")]
```
OliviaJSmith: appreciate the chuck norris jokes that actually made me laugh!

PR #48, opened 2025-10-14
Assignment 3

kellanburns:
Worked on with @BillMillerCoding .


CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
Hellothereyoko, line 1: You need to set this to the latest version to meet assignment requirements

CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
Hellothereyoko, line 1: I like how you combined the act & assert phases of testing. Do you have seperate test classes for each test type? How do they interact with the bot grade checker?

CanHazFunny/CanHazFunny/Program.cs
Hellothereyoko, line 1: Very simple main program, very clean. But I'd like to see some kind of user interaction so that they feel in control of the program...

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
Hellothereyoko, line 1: In order to avoid hyper specific imports; you can just use System to save yourself a couple of lines

Hellothereyoko (commented):
So far so good, missing a couple of things. Otherwise, a super strong build of the app. Well done!

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,90 @@
...
+    }
+
+    [Fact]
+    public void TellJoke_ChuckNorrisJoke_GeneratesNewJoke()
+    {
+        // Arrange
+        var mockOutput = new Mock<IOutput>();
+        var mockJokeService = new Mock<IJokeService>();
+        mockJokeService.SetupSequence(js => js.GetJoke())
+            .Returns("Norris counted to infinity. Twice.")
```
RyanHirte, line 65: Having Chuck or Norris in the joke is a good edge case to think about.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
RyanHirte: Actually I think they need to keep using System.IO as well as using System in order for these tests to compile. However, using CanHazFunny on line 2 I do think can be removed.

CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
```diff
@@ -0,0 +1,38 @@
+using Xunit;
+using CanHazFunny;
```
RyanHirte: I think using CanHazFunny can be removed here.

CanHazFunny/CanHazFunny.Tests/JokeServiceTests.cs
```diff
@@ -0,0 +1,23 @@
+using Xunit;
+using CanHazFunny;
```
RyanHirte: I think every import except using Xunit can be removed here.

RyanHirte (commented):
Create two interfaces. One to represent the JokeService and one interface to represent outputting the joke to the screen (Console.WriteLine). ✔

The JokeService will need to have the interface applied to it. ✔

Create an implementation of the output interface that writes out the joke to the console. ✔

Implement the Jester class. It should take in both interfaces as dependencies. These dependencies should be null checked. ✔

The Jester class TellJoke() method should retrieve a joke from the JokeService. If the joke contains "Chuck Norris", skip it and get another. The joke should be written to the output dependency. ✔

Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔

Nullability for all projects ✔

Ensure that you turn on treat warnings as errors for all projects ✔

Ensure that you turn on CodeAnalysisTreatWarningsAsErrors ✔

Ensure that you turn on EnforceCodeStyleInBuild ✔

Ensure there are no errors or warnings (including code analysis warnings) ✔

Overall great job other than some nitpicks I had with the using statements!

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
kellanburns: True, I chose to keep the imports so I could glance at the top of the file and know what was going on; it doesn't throw any errors, and seemed to help me move faster.

CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
kellanburns: *Didn't mean to resolve before commenting, adding comment now after*

I'm not sure how combining them effects any automatic grading, but the test code-coverage checker still picks it up and recognizes it, and I think this is much more efficient than anything else I can think of to accomplish the same result. 

The tests are also designed to have 1 test file (w/ 1 class) per appropriate project file, and all the tests for the file are in the single class. This is simpler than having multiple classes, and multiple classes (given the number of tests) seems unnecessary.

CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
kellanburns: Good catch, thanks for calling that out, I'm moving it up to `net9.0` now.

CanHazFunny/CanHazFunny/Program.cs
kellanburns: I didn't interpret any need for user input based on the assignment instructions, and while it could add functionality to the program, I don't think it would really improve its efficiency or fulfillment of the requirements.

CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
kellanburns: I mentioned above, I kept the extra imports for my understanding while developing, I could remove them now but I'm planning a refactor where I think I may use them.

CanHazFunny/CanHazFunny.Tests/JokeServiceTests.cs
kellanburns: ^ See above comment, I'm keeping them so I can understand while I'm looking through the code during development.

c-stanton:
For this assignment do the following:
•	Create two interfaces. One to represent the JokeService and one interface to represent outputting the joke to the screen (Console.WriteLine). ✔
•	The JokeService will need to have the interface applied to it. ✔
•	Create an implementation of the output interface that writes out the joke to the console. ✔
•	Implement the Jester class. It should take in both interfaces as dependencies. These dependencies should be null checked. ✔
•	The Jester class TellJoke() method should retrieve a joke from the JokeService. If the joke contains "Chuck Norris", skip it and get another. The joke should be written to the output dependency. ✔
•	Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔
Fundamentals
•	Be sure you enable: 
o	Nullability for all projects ✔
o	Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ❌(target framework is set to net8.0, which is not the latest release right now)
o	Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ✔
o	Ensure that you turn on treat warnings as errors for all projects ✔
o	Ensure that you turn on CodeAnalysisTreatWarningsAsErrors ✔
o	Ensure that you turn on EnforceCodeStyleInBuild ✔
•	Ensure there are no errors or warnings (including code analysis warnings) ✔
•	All of the above should be unit tested ✔
•	Choose simplicity over complexity ✔
Nice job!

CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
```diff
@@ -2,7 +2,12 @@
 
   <PropertyGroup>
     <TargetFramework>net8.0</TargetFramework>
```
c-stanton: Maybe change this to the latest version, net9.0

CanHazFunny/CanHazFunny/CanHazFunny.csproj
```diff
@@ -1,8 +1,15 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <OutputType>Exe</OutputType>
-    <TargetFramework>net8.0</TargetFramework>
-  </PropertyGroup>
+	<PropertyGroup>
+		<OutputType>Exe</OutputType>
+		<TargetFramework>net8.0</TargetFramework>
```
c-stanton: Same here, maybe change to net9.0

c-stanton (commented):
I would recommend changing these target frameworks to the latest version

CanHazFunny/CanHazFunny.Tests/CanHazFunny.Tests.csproj
kellanburns: Just fixed it

CanHazFunny/CanHazFunny/CanHazFunny.csproj
kellanburns: Just fixed it too

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+
+namespace CanHazFunny;
+
+public class Jester
+{
+    private readonly IOutput Output;
+    private readonly IJokeService JokeService;
```
Joshua-Lester3, line 8: [Guideline](https://essentialcsharp.com/guidelines#properties):
> DO favor automatically implemented properties over fields.

Joshua-Lester3 (commented):
## Assignment

For this assignment do the following:

- Create *two* interfaces. One to represent the `JokeService` and one interface to represent outputting the joke to the screen (`Console.WriteLine`). ✔
- The `JokeService` will need to have the interface applied to it. ✔
- Create an implementation of the output interface that writes out the joke to the console. ✔
- Implement the `Jester` class. It should take in both interfaces as dependencies. These dependencies should be null checked. ✔
- The `Jester` class `TellJoke()` method should retrieve a joke from the `JokeService`. If the joke contains "Chuck Norris", skip it and get another. The joke should be written to the output dependency. ✔
- Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔

## Fundamentals

- Be sure you enable:
  - Nullability for all projects  ✔
  - Set `LangVersion` and the `TargetFramework` to the latest released versions available (preview versions optional) ✔
  - Ensure that you turn on code analysis for all projects(`EnableNETAnalyzers`) ✔
  - Ensure that you turn on treat warnings as errors for all projects ✔
  - Ensure that you turn on `CodeAnalysisTreatWarningsAsErrors` ✔
  - Ensure that you turn on `EnforceCodeStyleInBuild` ✔
- **Ensure there are no errors or warnings (including code analysis warnings)** ✔
- **All of the above should be unit tested** ✔
- **Choose simplicity over complexity** ✔
  
## Extra Credit

- Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)? ✔
- The Geek jokes API that is being used can also return jokes in a JSON format. Update the `JokeService` to retrieve jokes using JSON. `GetJoke` should still return a string. ❌ Not implemented.

Overall, looks great guys! Good use of Moq, and setting/reverting Console.Out. Only comment was on properties as opposed to fields.

CanHazFunny/CanHazFunny/Jester.cs
OliviaJSmith: for properties that are only set in the constructor and not used outside of the class the way they are implemented here is standard but generally yes auto props are preferred - will double check with Ben on this to eliminate confusion

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,29 @@
...
+    public Jester(IOutput output, IJokeService jokeService)
+    {
+        ArgumentNullException.ThrowIfNull(output);
```
OliviaJSmith: since these parameters aren't nullable - don't need to throw if null here

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,29 @@
...
+        JokeService = jokeService;
+    }
+
+    public void TellJoke()
+    {
+        bool isChuckNorris = false;
+        string joke;
+        do {
+            joke = JokeService.GetJoke();
+            ArgumentNullException.ThrowIfNullOrWhiteSpace(joke);
```
OliviaJSmith, line 23: it's a bit redundant to throw for null/whitespace both here and in ShowOutput - if handling here then you don't need to handle in ShowOutput for this scenario

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,29 @@
...
+            isChuckNorris = joke.Contains("Chuck", StringComparison.OrdinalIgnoreCase)
```
OliviaJSmith, line 24: seems like we want to filter out specifically chuck norris jokes but this code would also filter out other chuck not norris jokes and same for checking Norris independantly here as well.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(output);
+        ArgumentNullException.ThrowIfNull(jokeService);
+        Output = output;
+        JokeService = jokeService;
+    }
+
+    public void TellJoke()
+    {
+        bool isChuckNorris = false;
```
OliviaJSmith, line 19: nit: general rule is to avoid extra variable assignment unless it's necessary. Since the while is the only place you'll have to check the value of this variable I'd say it's not needed here - this helps to avoid potential performance costs

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,89 @@
...
+            var jester = new Jester(new ConsoleOutput(), mockJokeService.Object);
+            jester.TellJoke();
+        }
+        finally
+        {
+            Console.SetOut(original);
+        }
+
+        // Assert
+        Assert.Contains(testJoke, sw.ToString());
```
OliviaJSmith, line 55: rather than contains you might want to test if they're equal?

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,89 @@
...
+            jester.TellJoke();
+        }
+        finally
+        {
+            Console.SetOut(original);
+        }
+
+        // Assert
+        string output = sw.ToString();
+        Assert.DoesNotContain("Chuck", output);
```
OliviaJSmith, line 85: nit: generally want one assert per test

CanHazFunny/CanHazFunny/IJokeService.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System;
+
+namespace CanHazFunny;
+
+public interface IJokeService
+{
+    public string GetJoke();
```
OliviaJSmith, line 7: generally would not put "public" or other on the interface. The implementation is what would set the method as public or private.

CanHazFunny/CanHazFunny.Tests/ConsoleOutputTests.cs
```diff
@@ -0,0 +1,37 @@
+using Xunit;
+using System;
+
+namespace CanHazFunny.Tests;
+
+public class ConsoleOutputTests
+{
+    [Fact]
+    public void ShowOutput_MessageIsNull_ArgumentNullException()
```
OliviaJSmith, line 9: These tests focus just on whether your exception is thrown while there is still the functionality of writing the output to the console and that is not directly tested here. It could be more clear to just assert that console output matches what you expect

PR #61, opened 2025-10-26
Assignment 5

chanderlud:
I completed this assignment with @ulises-aguilar 

We did the second extra credit option on this branch: https://github.com/chanderlud/EWU-CSCD371-2025-Fall/tree/Assignment5-ExtraCredit


GenericsHomework/Node.cs
```diff
@@ -0,0 +1,55 @@
...
+        };
+        Next = newNode;
+    }
+
+    public bool Exists(TNodeValue value)
+    {
+        Node<TNodeValue> current = this;
+        do
+        {
+            if (value is null && current.Value is null)
```
PigPig-SDK, line 32: I feel this if, else could be simplified.

PigPig-SDK (commented):
Define the Node class

Talked with Chandler, seems that 

JosephPotapenko:
Instructions mention: 
For this assignment, always use Assert.AreEqual<T>() (the generic version) 

But for the tests you ended up using Assert.Equal(). (if I understood it correctly):

so this:
Assert.Equal(0, node.Value);
Assert.Equal(node, node.Next);

would turn into (for some reason it gets rid of <.i.n.t.> if spelled without the periods, so I wrote it out the way I did to make it not dissapear, so remove the periods) : 
"Assert.Equal<.i.n.t.>(0, node.Value);"
"Assert.Equal<Node<.i.n.t.>>(node, node.Next);"

And so on and so forth. There's like 4 other instances where you used Assert.Equals() without the generic versions of the Assert.AreEqual<T>().

Otherwise, it's looking good to me (as far as I could tell).

JosephPotapenko:
## Instructions

- Create a *class library* project called "GenericsHomework.". ✔
  - Define the `Node` class ✔

## Fundamentals

- For this assignment, always use `Assert.AreEqual<T>()` (the generic version)  ❌

GenericsHomework/Node.cs
ulises-aguilar: Thanks for the feedback.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public void Clear()
+    {
+        // this is sufficient for garbage collection to clean up the nodes because there are no external references to them
+        // the nodes do not need to have their Next pointers set to null individually as long as there are no external references to them
+        Next = this;
+    }
+
+    override public string ToString()
```
Joshua-Lester3, line 51: nit: I'd do 'override' after 'public'. This still works/compiles, but the other way is more conventional.

Joshua-Lester3 (commented):
## Instructions

- Create a _class library_ project called "GenericsHomework.". ✔
  - Define the `Node` class

## Extra Credit

## Fundamentals

PR #63, opened 2025-10-27
Assignment 5

PigPig-SDK:
Worked with @ColtonKnopik


GenericHomework.Tests/GenericHomework.Tests.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+	<LangVersion>latest</LangVersion>
+	<EnableNETAnalyzers>true</EnableNETAnalyzers>
+	<AnalysisLevel>latest</AnalysisLevel>
+	<TreatWarningsAsErrors>false</TreatWarningsAsErrors>
+    <IsPackable>false</IsPackable>
```
KorbinWeiler: These should be in the Diretory.Build.props folder and shouldn't be needed here.

GenericHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        // Assert
+        Assert.AreEqual<Node<string>>(node.Next, node);
+    }
+
+    [TestMethod]
+    public void Node_Append_AppendsNext()
+    {
+        // Arrange
+        var node = new Node<string>("test1");
+        string test2 = "test2";
```
KorbinWeiler, line 71: Why assign a variable to take "test2" instead of just passing "test2" to the Node directly.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <summary>
+    /// Inserts a new node immediately after this node in the list.
+    /// </summary>
+    /// <remarks>
+    /// This method does <b>not</b> append to the end of the list.
+    /// Instead, it places the new node directly in front of this node’s current successor.
+    /// </remarks>
+    public void Append(T value)
+    {
+        if(Exists(value)) throw new ArgumentException("Given value is already included in this list!");
```
KorbinWeiler: I think this would be unnecessary since there's no requirement for each value in the list to be unique.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,52 @@
...
+        Node<T> node = new Node<T>(value);
+        node.Next = Next;
+        this.Next = node;
+    }
+    public void Clear()
+    {
+        //It is sufficent to set Next to 'this'
+        //Because GC is a tree structure, anything not connected to the 'root node' is eligible for cleanup.
+        //Therefore the rest of the list, now disconnected from the 'root node', will be cleaned up at some point in time.
+        Next = this;
```
KorbinWeiler, line 50: Would this not leave the root node intact. Like if I had a list 1->2->3 this would just make the list 1 but that first node wouldn't be cleaned up by the garbage collector because it is still in scope.

KorbinWeiler (commented):
Create a class library project called "GenericsHomework". ✔
Define the Node class
Extra Credit

Fundamentals

GenericsHomework/Node.cs
KorbinWeiler, line 39: disregard

GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+	<PropertyGroup>
+		<RootNamespace>GenericsHomework_</RootNamespace>
```
ulises-aguilar, line 3: This also shouldn't be included, as Korbin mentioned already.

ulises-aguilar (commented):
Define the Node class

kellanburns:
- Create a *class library* project called "GenericsHomework.". ❌
- Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item. ✔
  - Define the `Node` class ✔
  - The values should be homogeneous. ✔
  - Add a `Clear` method that effectively removes all items from a list except the current node. Pay attention as to whether you should be concerned with the following: ✔

Extra Credit

2. Implement `System.Collections.Generic.ICollection<T>` on the `Node` class ❌

Fundamentals

Super solid code, well written and my only fix is making the production project *specifically* a "class library" (which I only noticed because @chanderlud mentioned it in my project). Adding `<OutputType>Library</OutputType>` to the .csproj file should do the trick.

GenericsHomework/GenericsHomework.csproj
kellanburns, line 3: Adding this line should satisfy the first requirement to make it a "class library".

```suggestion
		<RootNamespace>GenericsHomework_</RootNamespace>
		<OutputType>Library</OutputType>
```

kellanburns (changes requested):

GenericsHomework/Node.cs
PigPig-SDK: I think that's the point of the the clear function. To clear all forward nodes in the list except the 'head' node.

GenericsHomework/GenericsHomework.csproj
PigPig-SDK: Thank you.

Joshua-Lester3 (commented):
## Instructions

- Create a _class library_ project called "GenericsHomework.". ✔
  - Define the `Node` class

## Extra Credit

## Fundamentals

PR #64, opened 2025-10-27
Assignment5

thigiang16:
Worked with @JosephPotapenko


GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
```
KorbinWeiler: These shouldn't be needed since they are already in your Directory.Build.props file.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,105 @@
...
+using System;
+using GenericsHomework;
+
+namespace GenericsHomework.Tests;
+
+    [TestClass]
+    public sealed class NodeTests
+    {
+        [TestMethod]
+        public void Constructor_ShouldInitializeValueandSelfReference()
```
KorbinWeiler: A better name for this test would something like Constructor_ValidInitialization_SelfReferences. This gives you the parts of method under test, condition of the test and expected results.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,105 @@
...
+        public void Append_DuplicateValue_ThrowsInvalidOperationException()
+        {
+            var node = new Node<int>(1);
+            node.Append(2);
+
+            Assert.ThrowsException<InvalidOperationException>(() => node.Append(2));
+        }
+
+        [TestMethod]
+        public void Exists_FindsValuesCorrectly()
```
KorbinWeiler: Doesn't follow the proper test naming. Could be "Exists_ExistingValue_FindsValue"

KorbinWeiler: You have a test for finding an existing value but you should also have a test that tries to find a value that is not in the list.

KorbinWeiler (commented):
Define the Node class
Extra Credit

Fundamentals

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <summary>
+        /// Removes all nodes from the circular list except the current node.
+        /// After this call the current node will point to itself (a single-node circular list).
+        /// </summary>
+
+        //Note: Garbage Collection in .NET is automatic and can detect, clean up cyclic references
+        //Therefore, although the nodes form a circular list,
+        //we don't need to manually break the loop for memory management
+
+        public void Clear()
```
steeley21, line 35: Small note, but the comments on the Clear() are safe to be removed. I don't think they are necessary to understand what this method is doing.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            Node<T> current = this;
+            do
+            {
+                if (object.Equals(current.Value, value))
+                    throw new InvalidOperationException("Duplicate value detected");
+                current = current.Next;
+            }while (current != this);
+
+            var newNode = new Node<T>(value) { Next = this.Next };
```
steeley21, line 31: Nitpick, but I would avoid using var here just for clarity. Instead:

Node<T> newNode = new Node<T>(value) { Next = this.Next };

steeley21 (commented):
Instructions

Define the Node class

Extra Credit

Fundamentals

GenericsHomework.Tests/GenericsHomework.Tests.csproj
JosephPotapenko: 👍

GenericsHomework.Tests/NodeTests.cs
JosephPotapenko: 👍

GenericsHomework.Tests/NodeTests.cs
JosephPotapenko: 👍

GenericsHomework/Node.cs
JosephPotapenko: 👍

GenericsHomework.Tests/NodeTests.cs
JosephPotapenko: 👍Nice catch. Don't know why I didn't notice that.

GenericsHomework/Node.cs
JosephPotapenko: Took me way too long to figure out that <.i.n.t.> disappears in comments 😂. Implemented 👍

GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
```
Joshua-Lester3, line 4: nit: I'd remove this, since it's specified in the root dir props file.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        public void Append(T value)
+        {
+            Node<T> current = this;
+            do
+            {
+                if (object.Equals(current.Value, value))
+                    throw new InvalidOperationException("Duplicate value detected");
+                current = current.Next;
+            }while (current != this);
```
Joshua-Lester3, line 29: This could be refactored to use the current Exists method.

Joshua-Lester3 (commented):
## Instructions

- Create a _class library_ project called "GenericsHomework.". ❌ Not specified as a library.
  - Define the `Node` class
    - Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and, therefore, may never be garbage collected. ❌ No comment.

## Extra Credit

## Fundamentals

GenericsHomework/Node.cs
thigiang16: Nice catch!
Btw, we have our Assignment 5 extra credit on a separate branch #68 , as Ben mentioned earlier in class. Could you recheck it when you have a chance?

PR #65, opened 2025-10-27
Assignment 5

katmilton:
I worked with @steeley21. We implemented both Extra Credit options.

### DIFF
--- a/GenericsHomework.Tests/GenericsHomework.Tests.csproj
+++ b/GenericsHomework.Tests/GenericsHomework.Tests.csproj
@@ -1,0 +1,25 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="coverlet.collector" Version="6.0.2" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.12.0" />
+    <PackageReference Include="xunit" Version="2.9.2" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.8.2" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\GenericsHomework\GenericsHomework\GenericsHomework.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Using Include="Xunit" />
+  </ItemGroup>
+
+</Project>
--- a/GenericsHomework.Tests/GenericsHomework.Tests.csproj
+++ b/GenericsHomework.Tests/GenericsHomework.Tests.csproj
@@ -18,8 +18,10 @@
     <ProjectReference Include="..\GenericsHomework\GenericsHomework\GenericsHomework.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
-    <Using Include="Xunit" />
-  </ItemGroup>
+
+	<ItemGroup>
+		<PackageReference Include="MSTest.TestFramework" Version="3.*" />
+		<PackageReference Include="MSTest.TestAdapter" Version="3.*" />
+	</ItemGroup>
 
 </Project>
--- a/GenericsHomework.Tests/VennDiagramTests.cs
+++ b/GenericsHomework.Tests/VennDiagramTests.cs
@@ -23,10 +23,27 @@ public class VennDiagramTests
     [TestMethod]
     public void Union_ReturnsAllUniqueItems_Success()
     {
+        var vd = new VennDiagram<string>();
+        var a = vd.AddCircle("A");
+        var b = vd.AddCircle("B");
+        a.Add("x"); a.Add("y");
+        b.Add("y"); b.Add("z");
+
+        var union = vd.Union("A", "B").OrderBy(s => s).ToList();
+        CollectionAssert.AreEqual(new[] { "x", "y", "z" }, union);
     }
 
     [TestMethod]
     public void Difference_ReturnsOnlyLeftMinusRightItems_Success()
     {
+        var vd = new VennDiagram<string>();
+        var a = vd.AddCircle("A");
+        var b = vd.AddCircle("B");
+        a.Add("x"); a.Add("y");
+        b.Add("y"); b.Add("z");
+
+        var diff = vd.Difference("A", "B").ToList();
+        Assert.AreEqual<int>(1, diff.Count);
+        Assert.AreEqual<string>("x", diff[0]);
     }
 }
--- a/GenericsHomework.Tests/VennDiagramTests.cs
+++ b/GenericsHomework.Tests/VennDiagramTests.cs
@@ -14,8 +14,10 @@ public class VennDiagramTests
         var vd = new VennDiagram<string>();
         var a = vd.AddCircle("A");
         var b = vd.AddCircle("B");
-        a.Add("x"); a.Add("y");
-        b.Add("y"); b.Add("z");
+        a.Add("x"); 
+        a.Add("y");
+        b.Add("y"); 
+        b.Add("z");
 
         var intersection = vd.Intersection("A", "B").ToList();
         Assert.AreEqual<int>(1, intersection.Count);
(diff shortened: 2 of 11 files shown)


GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
```
RyanHirte, line 4: You don't need ton include some of this since it is passed down from Directory.Build.props, but either way it works.

GenericsHomework/GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
```
RyanHirte, line 4: Same thing here with not needing some of this because it's in Directory.Build.props

GenericsHomework/GenericsHomework/VennDiagram.cs
```diff
@@ -0,0 +1,50 @@
...
+	public bool Contains(T item) => _items.Contains(item);
+	public IEnumerable<T> Items => _items;
+    }
+
+public class VennDiagram<T> where T: class
+{
+	private readonly Dictionary<string, Circle<T>> _circles = new();
+	public IEnumerable<Circle<T>> Circles => _circles.Values;
+
+	public Circle<T> AddCircle(string name)
```
RyanHirte, line 23: Unsure why AddCircle is returning a circle instead of a bool or something to indicate that adding the circle was a success, or making it a void method and not return anything. I get that it could potentially be used for testing, but then why not just use the Get method below instead. Not saying it's wrong, maybe I'm missing something but I'm a little confused on this.

GenericsHomework.Tests/VennDiagramTests.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    private static readonly string[] ExpectedUnion = { "x", "y", "z" };
+
+    [TestMethod]
+    public void Intersection_ReturnsOnlyCommonItems_Success()
+    {
+        var vd = new VennDiagram<string>();
+        var a = vd.AddCircle("A");
+        var b = vd.AddCircle("B");
+        a.Add("x"); a.Add("y");
```
RyanHirte: I haven't seen code formatted like this before, maybe put these on their own line? It doesn't change anything but it could make it slightly more readable.

RyanHirte (commented):
Define the Node class

Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ✔

Great job! Only thing I would note is that I think it would be better if the extra credit were in it's own separate branch like we talked about in class before, that way it is easier to distinguish from the actual assignment, but I can't remember if that was a requirement or just preferred. Nice work!

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,227 @@
...
+    private bool _isEmpty;
+
+    /// <summary>
+    /// Initializes a single-node circular list; Next points to this node.
+    /// _isEmpty will represent an empty collection while keeping the self-loop.
+    /// </summary>
+    public Node(T value)
+    {
+        _value = value;
+        _isEmpty = false;
```
c-stanton: Just a nitpick, but try to avoid naming fields with camel case

c-stanton:
•	Create a class library project called "GenericsHomework.". ✔
•	Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item. 
o	Define the Node class
o	The values should be homogeneous.
o	Include a constructor that takes a value. (No validation is necessary on the value). ✔
o	Add a ToString() override that writes out the value's ToString() result. ✔
o	Add a Next property that references the next node or else refers back to itself if there are no other nodes in the list. ✔ 
	The Next property should be non-nullable (careful to follow the non-nullable property guidelines) ✔
	The Next property setter should be private. ✔
o	Add an Append method that takes a value and appends a new Node instance after the current node (by invoking the Next property). ✔
o	Add a Clear method that effectively removes all items from a list except the current node. Pay attention as to whether you should be concerned with the following: 
	Whether it is sufficient to only set Next to itself ✔
	Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ✔
	Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and, therefore, may never be garbage collected. ✔
o	Create an Exists method to test whether a value exists in the list. ✔
o	Throw an error on an attempt to Append a duplicate value. (Make sure you test for this case) ✔
•	You should not rely on any BCL generic classes for your implementation. ✔

    Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ✔

Fundamentals
•	Place all shared project properties into a Directory.Build.Props file. ✔
•	nullable reference types is enabled ✔
•	Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ✔
•	Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔
•	and enabled .NET analyzers for both projects ✔
•	For this assignment, always use Assert.AreEqual<T>() (the generic version) ✔
•	All of the above should be unit tested ✔
•	Choose simplicity over complexity ✔

Nice work!

Hellothereyoko:
Define the Node class

For this assignment, always use Assert.AreEqual() (the generic version) ✔

Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ✔
Implement System.Collections.Generic.ICollection on the Node class ✔



Everything required is all here. But, pay attention to your code coverage. There's some cases your code isn't catching

GenericsHomework.Tests/VennDiagramTests.cs
steeley21: Moved these to their own lines for better readability. Thanks!

GenericsHomework/GenericsHomework/Node.cs
steeley21: Thanks for catching that. Just updated.

GenericsHomework/GenericsHomework/VennDiagram.cs
katmilton: By returning a Circle we get immediate access to it as well. So if I wanted to add an item to the circle immediately I could. In addition, it assures that when calling AddCircle("A") twice, I am receiving the same circle reference. I also don't have to call a separate get or check for null.

Thank you for the feedback!

GenericsHomework.Tests/GenericsHomework.Tests.csproj
steeley21: Yes, agreed this isn't strictly necessary. Leaving it for now since it doesn’t affect anything. Thanks!

GenericsHomework/GenericsHomework/GenericsHomework.csproj
steeley21: Going to leave this for now since it isn't strictly necessary to remove. Thanks!

GenericsHomework.Tests/GenericsHomework.Tests.csproj
Joshua-Lester3: I'd agree with Ryan - it's redundant keeping it in. It makes me think you're overriding it, but they're the same so a bit confusing.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        this.Next = this;
+        _IsEmpty = true;
+
+        // Garbage Collector note:
+        // If nothing outside the linked list still references the removed nodes,
+        // they’ll be collected even if they point to each other.
+        // Using a self-loop ensures a removed node can’t point back into the list
+    }
```
Joshua-Lester3, line 91: All of this is unnecessary. As long as there are no external references to the potentially garbage collected nodes, they will get cleaned up. So those last two lines of code are all that's needed. Plus I would refrain from using `this.`.

GenericsHomework/GenericsHomework/VennDiagram.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+	public string Name { get; }
+	private readonly HashSet<T> _items = new();
+
+	public Circle(string name) => Name = name;
+	public bool Add(T item) => _items.Add(item);
+	public bool Contains(T item) => _items.Contains(item);
+	public IEnumerable<T> Items => _items;
+    }
+
```
Joshua-Lester3, line 17: There's a guideline to put classes in separate files:
> AVOID placing more than one class in a single source file.

Joshua-Lester3 (commented):
## Instructions

- Create a _class library_ project called "GenericsHomework.". ❌ Not specified as a library.
  - Define the `Node` class

## Extra Credit

## Fundamentals

PR #66, opened 2025-10-28
Assignment5

BillMillerCoding:
my partner: @KorbinWeiler
Link to extra credit PR: https://github.com/KorbinWeiler/EWU-CSCD371-2025-Fall/pull/2

### DIFF
--- a/GenericsHomework/Node.cs
+++ b/GenericsHomework/Node.cs
@@ -3,17 +3,17 @@
 public class Node<T>
 {
     public T Data { get; set; }
-    private Node<T> next;
+    private Node<T> _next;
     public Node<T> Next
     {
-        get => next;
-        private set => next = value ?? this;
+        get => _next;
+        private set => _next = value ?? this;
     }
 
     public Node(T data)
     {
         Data = data;
-        next = this;
+        _next = this;
     }
 
     public void Append(T value)
--- a/GenericsHomework/Node.cs
+++ b/GenericsHomework/Node.cs
@@ -1,73 +1,72 @@
-﻿namespace GenericsHomework
+﻿namespace GenericsHomework;
+
+public class Node<T>
 {
-    public class Node<T>
+    public T Data { get; set; }
+    private Node<T> _next;
+    public Node<T> Next
     {
-        public T Data { get; set; }
-        private Node<T> _next;
-        public Node<T> Next
-        {
-            get => _next;
-            private set => _next = value ?? this;
-        }
+        get => _next;
+        private set => _next = value ?? this;
+    }
 
-        public Node(T data)
-        {
-            Data = data;
-            _next = this;
-        }
+    public Node(T data)
+    {
+        Data = data;
+        _next = this;
+    }
 
-        public void Append(T value)
+    public void Append(T value)
+    {
+        if (this.Exists(value))
         {
-            if (this.Exists(value))
-            {
-                throw new InvalidOperationException("Cannot Have Duplicates Appended");
-            }
-            Node<T> nextNode = new(value)
-            {
-                Next = this.Next
-            };
-            Next = nextNode;
+            throw new InvalidOperationException("Cannot Have Duplicates Appended");
         }
+        Node<T> nextNode = new(value)
+        {
+            Next = this.Next
+        };
+        Next = nextNode;
+    }
 
-        public void Clear()
+    public void Clear()
+    {
+        Next = this;
+    }
+    public void IterativeClear()
+    {
+        Node<T> cur = this.Next;
+        while (cur != this)
         {
-            Next = this;
+            Node<T> next = cur.Next;
+            cur.Next = cur;
+            cur = next;
         }
-        public void IterativeClear()
-        {
-            Node<T> cur = this.Next;
-            while (cur != this)
-            {
-                Node<T> next = cur.Next;
-                cur.Next = cur;
-                cur = next;
-            }
-            this.Next = this;
-        }
+        this.Next = this;
+    }
 
-        public bool Exists(T value)
+    public bool Exists(T value)
+    {
+        Node<T> cur = this;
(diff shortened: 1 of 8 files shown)


BillMillerCoding:
I had to remove all underscores from test method names because it was failing on the shared build and test.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,72 @@
...
+            throw new InvalidOperationException("Cannot Have Duplicates Appended");
+        }
+        Node<T> nextNode = new(value)
+        {
+            Next = this.Next
+        };
+        Next = nextNode;
+    }
+
+    public void Clear()
```
steeley21, line 34: The Clear() method is missing a comment about garbage collection. 

From the requirements:

Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and, therefore, may never be garbage collected. ❌

GenericsHomeworkTests/UnitTest1.cs
```diff
@@ -0,0 +1,220 @@
+﻿using GenericsHomework;
+using System.Security.Cryptography.X509Certificates;
+using Xunit;
+
+namespace GenericsHomeworkTests;
+
+public class UnitTest1
```
steeley21: The naming on this test should match the Node class, to show that's the class it's testing. Something like NodeTests.cs

GenericsHomeworkTests/NodeTests.cs
```diff
@@ -0,0 +1,220 @@
...
+using Xunit;
+
+namespace GenericsHomeworkTests;
+
+public class UnitTest1
+{
+    public record class TestClass(string Name, int Id);
+
+    [Fact]
+    public void NameSingleNodeLinkedListNextIsSetToSelf()
```
thigiang16, line 12: All of your test names don’t really follow the proper naming. You could rename them to make each test’s purpose clearer

steeley21 (commented):
Define the Node class
Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and, therefore, may never be garbage collected. ❌

Extra Credit

Fundamentals

GenericsHomework/Node.cs
thigiang16, line 34: Missing a comment to indicate whether you need to worry about garbage collection

thigiang16 (commented):
Define the Node class
Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and, therefore, may never be garbage collected. ❌
Extra Credit

Fundamentals

GenericsHomework/Node.cs
BillMillerCoding: Fixed!

GenericsHomeworkTests/NodeTests.cs
BillMillerCoding: In this case, i couldnt add underscores because it was failing the shared test build even though it should be legal.

GenericsHomework/Node.cs
BillMillerCoding: Added

GenericsHomeworkTests/UnitTest1.cs
BillMillerCoding: Fixed!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,74 @@
...
+    public Node<T> Next
+    {
+        get => _next;
+        private set => _next = value ?? this;
+    }
+
+    public Node(T data)
+    {
+        Data = data;
+        _next = this;
```
Joshua-Lester3, line 16: You should always access the backing field through the property. Guideline:
> AVOID accessing the backing field of a property outside the property, even from within the containing class.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,74 @@
+﻿namespace GenericsHomework;
+
+public class Node<T>
+{
+    public T Data { get; set; }
+    private Node<T> _next;
+    public Node<T> Next
+    {
+        get => _next;
+        private set => _next = value ?? this;
```
Joshua-Lester3, line 10: This doesn't make a lot of sense to me. Say this Node is pointing to another Node linked to a few more. So if someone tries to append a null Node type, this would functionally clear the list. I'm not against the null check, but I think it could be done a different, better way.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+        Node<T> nextNode = new(value)
+        {
+            Next = this.Next
+        };
+        Next = nextNode;
+    }
+    //This Clear() is not as good as itterative clear because the garbage collector wont collect
+    //any of the nodes that point to the node clear is called on. It is also bad practice because
+    //Some node's next can access the node cleared is called on.
```
Joshua-Lester3, line 33: I'm pretty sure (@BenjaminMichaelis correct if wrong) that they will be garbage collected with the following Clear method. Since there are no external references to the potentially collected nodes, they will get collected.

GenericsHomeworkTests/NodeTests.cs
Joshua-Lester3: I'm confused by this as well. What was failing?

Joshua-Lester3 (commented):
## Instructions

- Create a _class library_ project called "GenericsHomework.". ❌ Not specified.
  - Define the `Node` class

## Extra Credit

1. Implement a `VennDiagram` structure that contains `n` `Circle`s that only contains homogenous **reference types** of any type. ❌ not building/passing Actions pipeline

1. Implement `System.Collections.Generic.ICollection<T>` on the `Node` class ❌ not implemented.

## Fundamentals

PR #67, opened 2025-10-28
Assignment 5

c-stanton:
Worked with @Al3xramirez and added in extra credits for both ICollection<T> on Node class and the VennDiagram structure.


GenericsHomeWork.Tests/GenericsHomeWork.Tests.csproj
```diff
@@ -0,0 +1,28 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <EnableNetAnalyzers>true</EnableNetAnalyzers>
```
ulises-aguilar: Place all shared project properties into a Directory.Build.Props file.

JosephPotapenko:
Was about to say the same thing Ulises said. No Directory.Build.Props to fulfill the requirement:

ulises-aguilar (commented):
Define the Node class

JosephPotapenko:
You could add in checks in the add and contains methods to make sure the item passed in isn't null, something like this possibly:
Contains:
     return item != null && _items.Contains(item);
Add:
    ArgumentNullException.ThrowIfNull(item);
    _items.Add(item);

You could check that circle name doesn't come in null like Mark showed in class yesterday:
    Name = name ?? throw new ArgumentNullException(nameof(name));


Once you set up  Directory.Build.props from my previous comment, you can get rid of the duplicate code from GenericsHomeWork.csproj and GenericsHomeWork.Tests.csproj. 


Possibly do some null checks on the extra credit too if you want?

Most of my nitpicks are edge cases though, you seem to have gotten most everything else done well.

JosephPotapenko:
## Instructions

- Create a *class library* project called "GenericsHomework.". ✔
  - Define the `Node` class

## Extra Credit

It looks like it was attempted. Didn't check it since it's not specifically required.

## Fundamentals

- Place all shared project properties into a `Directory.Build.Props` file. ❌

GenericsHomeWork.Tests/GenericsHomeWork.Tests.csproj
Al3xramirez: Thank you Ulises, will get that fixed

kellanburns:
- Create a class library project called "GenericsHomework.". ❌ (not a 'class library')
- Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item. ✔
   - Define the Node class ✔
   - The values should be homogeneous. ✔
   - Add a ToString() override that writes out the value's ToString() result. ✔
   - Add a Next property that references the next node or else refers back to itself if there are no other nodes in the list. ✔
   - Add an Append method that takes a value and appends a new Node instance after the current node (by invoking the Next property). ✔
   - Add a Clear method that effectively removes all items from a list except the current node. Pay attention as to whether you should be concerned with the following:

Extra Credit
1. Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ✔
- Each circle contains n items and each item can belong to one or more Circle instances.
- You are not required to use a Node from earlier in the homework for your Venn diagram implementation.
2. Implement System.Collections.Generic.ICollection<T> on the Node class ✔

Fundamentals
- Place all shared project properties into a Directory.Build.Props file.
nullable reference types is enabled ❌ (no .Props file)
- For this assignment, always use Assert.AreEqual<T>() (the generic version) ✔

Most of it looks good, previous reviews have already mentioned the issues with 'class library' and the missing Directory.Build.Props file. I also think the exception thrown by `Append()` should be an `ArgumentException`, because the provided `InvalidOperationException` is more relevant to an object's state than a generally disallowed operation (which more closely fits the parameters of an ArgumentException).

GenericsHomeWork/Node.cs
```diff
@@ -0,0 +1,140 @@
...
+    public bool Contains(T item)
+    {
+        return Exists(item);
+    }
+
+    public void Append(T value)
+    {
+        if (Exists(value))
+        {
+            throw new InvalidOperationException($"Value '{value}' already exists in the list.");
```
kellanburns: I think the exception type should be changed from `InvalidOperationException` to `ArgumentException`, because the use-case for the InvalidOperation exception is when the current state of an object makes the operation invalid, not that the operation isn't allowed to happen. 

Also, a bit of a nitpick, formatting the exception with `nameof(value)` being passed after the message is more common and lets you describe the 'no duplicates allowed', versus just alerting about the duplicate. 

```suggestion
            throw new ArgumentException("No duplicate values allowed in the list.", nameof(value));
```

kellanburns (changes requested):

c-stanton:
> * s

It defaults to a class library when there is an absence of a Main. The build gives it a .dll extension, and when I view the properties of the project it shows that it is a class library: 
default
generics1
generics2

GenericsHomeWork/Node.cs
c-stanton: I changed it, thank you for your suggestion!

katmilton (commented):
Define the Node class


Extra Credit

Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ✔


Fundamentals

Great work!

GenericsHomeWork/Node.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+        this.Next = this;
+    }
+
+    public int Count
+    {
+        get
+        {
+            int count = 0;
+            NodeCollection<T>? currNode = this.Next;
```
Joshua-Lester3, line 74: I'm not sure why this is a nullable variable, since Next is non-nullable. Also, I'd avoid using `this.` as it's not needed here.

Joshua-Lester3 (commented):
## Instructions

- Create a _class library_ project called "GenericsHomework.". ❌ Not specified in csproj
  - Define the `Node` class

## Extra Credit

## Fundamentals

Joshua-Lester3:
> It defaults to a class library when there is an absence of a Main. The build gives it a .dll extension, and when I view the properties of the project it shows that it is a class library:

I think it defaults to this if there's no Main, but specifying it explicitly avoids ambiguity.

PR #72, opened 2025-10-28
Assignment 5

RyanHirte:
Worked on with @Hellothereyoko

Extra credit PR: https://github.com/Hellothereyoko/EWU-CSCD371-2025-Fall/pull/13


GenericsHomework/Node.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// <summary>
+    /// Appends a new node with the specified value to the circular list.
+    /// Throws ArgumentException if the value already exists in the list.
+    /// </summary>
+    public void Append(T value)
+    {
+        // Check for duplicates before appending
+        if (Exists(value))
+        {
+            throw new ArgumentException($"Value '{value}' already exists in the list.", nameof(value)); //ERR CONDITION
```
ColtonKnopik: This should be changed to an `InvalidOperationException` because it is not the argument that is causing the exception as much as the operation being performed.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// </summary>
+    public bool Exists(T value)
+    {
+        Node<T> current = this;
+
+        // Check first node
+        if (EqualityComparer<T>.Default.Equals(current.Value, value))
+        {
+            return true;
+        }
```
ColtonKnopik: If you replace this loop with a do-while, you could remove the check for the first node.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,102 @@
...
+
+    public Node(T value) //Instantiate values for best access practice 
+    {
+        Value = value;
+        Next = this;
+    }
+
+    public override string ToString()
+    {
+        return Value?.ToString() ?? String.Empty; //checks if val present: if not, it returns an empty string to be handled
```
ColtonKnopik: This function would be more informative if instead of replacing the null string with `String.Empty` you replace it with `"null"`

ColtonKnopik (commented):
        Define the Node class

Good job

PigPig-SDK:
- Technically removed the period at the end. I expect this was a typo on the assignment.

Define the Node class
- This part of the assignment is poorly worded. While your solution does append a node 'after' the current node, its not appending nodes DIRECTLY after. What is intended I wouldn't know, so it gets a checkmark.

Fundamentals
- Uses Assert.Equal, should be good enough.

Colton covered the problems I saw.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,575 @@
...
+    [Fact]
+    public void Append_DuplicateValue_ThrowsArgumentException()
+    {
+        // Arrange
+        Node<int> firstNode = new Node<int>(1);
+        firstNode.Append(2);
+        firstNode.Append(3);
+
+        // Act & Assert
+        ArgumentException ex = Assert.Throws<ArgumentException>(() => firstNode.Append(2));
```
thigiang16: I think use InvalidOperationException for the Node class and the test would be better because the error is caused by to the operation, not an invalid argument

thigiang16 (approved):
Define the Node class
Extra Credit

Fundamentals

I also checked out your Extra Credit. It looks good!

Hellothereyoko:
Added Feedback needed

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,575 @@
...
+        public string Name { get; set; } = string.Empty;
+        public int Age { get; set; }
+
+        public override string ToString()
+        {
+            return $"{Name} ({Age})";
+        }
+    }
+
+    //Structure acts as a psuedoconstructor
```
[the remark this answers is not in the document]
RyanHirte: Not there anymore.

GenericsHomework/Node.cs
RyanHirte: Agreed, will implement.

GenericsHomework.Tests/NodeTests.cs
RyanHirte: This will be implemented.

GenericsHomework/Node.cs
RyanHirte: Thank you for the comment, but I actually think keeping it an empty string makes more sense because I really don't see a point where a programmer would want a node with no value, and I think keeping it an empty string more represents the null-ness to me. That way if we are checking for emptiness or null-ness, I think it's better to check for empty string than a string that says "null". I think this is more subjective though, not saying you are wrong in this, I just can't see it for this assignment.

GenericsHomework/Node.cs
RyanHirte: Much cleaner this will be implemented.

Hellothereyoko:
> Create a class library project called "GenericsHomework.". ✔ Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item. Define the Node class The values should be homogeneous. Include a constructor that takes a value. (No validation is necessary on the value). ✔ Add a ToString() override that writes out the value's ToString() result. ✔ Add a Next property that references the next node or else refers back to itself if there are no other nodes in the list. ✔ The Next property should be non-nullable (careful to follow the non-nullable property guidelines) ✔ The Next property setter should be private. ✔ Add an Append method that takes a value and appends a new Node instance after the current node (by invoking the Next property). ✔ Add a Clear method that effectively removes all items from a list except the current node. Pay attention as to whether you should be concerned with the following: Whether it is sufficient to only set Next to itself ✔ Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ✔ Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and, therefore, may never be garbage collected. ✔ Create an Exists method to test whether a value exists in the list. ✔ Throw an error on an attempt to Append a duplicate value. (Make sure you test for this case) ✔ You should not rely on any BCL generic classes for your implementation. ✔ Extra Credit Do one of the following two options (or both if you want extra, extra credit) :)
> 
> Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ❌ Each circle contains n items and each item can belong to one or more Circle instances. You are not required to use a Node from earlier in the homework for your Venn diagram implementation. You are welcome to use existing BCL generic classes for the extra credit. Implement System.Collections.Generic.ICollection on the Node class ✔ Fundamentals Place all shared project properties into a Directory.Build.Props file. nullable reference types is enabled ✔ Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ✔ Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔ and enabled .NET analyzers for both projects ✔ For this assignment, always use Assert.AreEqual() (the generic version) ✔ All of the above should be unit tested ✔ Choose simplicity over complexity ✔
> 
> I also checked out your Extra Credit. It looks good!

Thanks for your feedback! We went back and made some edits :)

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,980 @@
...
+        // Arrange
+        NodeCollection<int> firstNode = new NodeCollection<int>(1);
+
+        // Act
+        firstNode.Append(2);
+
+        // Assert
+        Assert.AreEqual<int>(2, firstNode.Next.Value);
+        Assert.AreSame(firstNode, firstNode.Next.Next);
+    }
```
Joshua-Lester3, line 136: These tests have the same content. I'd remove one of them.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,980 @@
...
+
+        // Assert
+        Assert.AreEqual<int>(1, firstNode.Value);
+        Assert.AreEqual<int>(2, firstNode.Next.Value);
+        Assert.AreEqual<int>(3, firstNode.Next.Next.Value);
+        Assert.AreSame(firstNode, firstNode.Next.Next.Next);
+    }
+
+    [TestMethod]
+    public void Append_MultipleNodes_MaintainsCircularStructure()
```
Joshua-Lester3, line 156: imo, this test is redundant, with the previous one in mind.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,980 @@
...
+    [TestMethod]
+    public void Append_DuplicateString_ThrowsInvalidOperationException()
+    {
+        // Arrange
+        NodeCollection<string> firstNode = new NodeCollection<string>("test");
+        firstNode.Append("hello");
+
+        // Act & Assert
+        Assert.ThrowsException<InvalidOperationException>(() => firstNode.Append("test"));
+    }
```
Joshua-Lester3, line 229: There's some other instances of this as well, but I think at this point you're just testing that C# generics work. So it's more of a test of the language, which I wouldn't include.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,980 @@
...
+        NodeCollection<int> firstNode = new NodeCollection<int>(1);
+        firstNode.Append(2);
+        firstNode.Append(3);
+
+        // Act
+        bool result = firstNode.Exists(3);
+
+        // Assert
+        Assert.IsTrue(result);
+    }
```
Joshua-Lester3, line 291: Redundant imo. Could also be refactored into a Data Driven Test.

Joshua-Lester3 (commented):
## Instructions

- Create a _class library_ project called "GenericsHomework.". ❌ Not specified
  - Define the `Node` class
  - Add an `Append` method that takes a value and appends a new `Node` instance after the current node (by invoking the `Next` property). ❌ Inserts after last node, not the current node

## Extra Credit

1. Implement a `VennDiagram` structure that contains `n` `Circle`s that only contains homogenous **reference types** of any type. ❌ not implemented

## Fundamentals

PR #73, opened 2025-10-28
Assignment 5

kellanburns:
No partner, I worked solo on this assignment.

### DIFF
--- a/GenericsHomework.Tests/GenericsHomework.Tests.csproj
+++ b/GenericsHomework.Tests/GenericsHomework.Tests.csproj
@@ -1,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
+    <Nullable>enable</Nullable>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="MSTest" Version="4.0.0-preview.25415.9" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Using Include="Microsoft.VisualStudio.TestTools.UnitTesting" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\GenericsHomework\GenericsHomework.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/GenericsHomework.Tests/GenericsHomework.Tests.csproj
+++ b/GenericsHomework.Tests/GenericsHomework.Tests.csproj
@@ -7,6 +7,7 @@
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <ImplicitUsings>enable</ImplicitUsings>
+    <NoWarn>$(NoWarn);MSTEST0037</NoWarn> <!-- Allows me to exclusively use Assert.AreEqual<T>() -->
   </PropertyGroup>
 
   <ItemGroup>
@@ -14,7 +15,8 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="MSTest" Version="4.0.0-preview.25415.9" />
+    <PackageReference Include="MSTest.TestAdapter" Version="4.0.1" />
+    <PackageReference Include="MSTest.TestFramework" Version="4.0.1" />
   </ItemGroup>
 
   <ItemGroup>
--- a/GenericsHomework.Tests/GenericsHomework.Tests.csproj
+++ b/GenericsHomework.Tests/GenericsHomework.Tests.csproj
@@ -7,7 +7,6 @@
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <ImplicitUsings>enable</ImplicitUsings>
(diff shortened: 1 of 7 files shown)


GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,35 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
```
RyanHirte, line 3: You actually don't need to repeat this project configuration because it is passed down from your Directory.Build.props file.

GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
```
RyanHirte, line 4: You actually don't need to repeat this project configuration because it is passed down from your Directory.Build.props file.

RyanHirte (commented):
Define the Node class

Extra credit not done.

Overall great job!

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+public class Node<T>
+{
+    private Node<T> _next;
+    public T Value { get; }
+
+    public Node<T> Next
+    {
+        get => _next;
+        private set => _next = value ?? throw new ArgumentNullException(nameof(value));
+    }
```
Al3xramirez, line 15: I don't think we need the "throw new Arguement" check as Next is set to private so you always set _next = this in the constructor. Correct me if im wrong!

Al3xramirez (commented):
Instructions
Define the Node class

For this assignment, always use Assert.AreEqual() (the generic version) ✔
Awesome work!

GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
chanderlud, line 3: I think you should add `<OutputType>Library</OutputType>` inside the PropertyGroup to meet the requirement `Create a *class library* project called "GenericsHomework.".`

chanderlud (commented):
- Create a *class library* project called "GenericsHomework.". ❌
  - Define the `Node` class

## Extra Credit

## Fundamentals

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace GenericsHomework;
+
+public class Node<T>
```
chanderlud, line 6: I would recommend renaming `T` to something more descriptive like `TNodeValue` based on the guideline `DO choose meaningful names for type parameters and prefix the name with T.`

BillMillerCoding:
Create a class library project called "GenericsHomework.". ❌
Define the Node class
Extra Credit

Fundamentals

GenericsHomework.Tests/GenericsHomework.Tests.csproj
kellanburns: Fair enough, but I suppose I'll just keep it for the sake of being explicit and for easier code readability.

GenericsHomework/GenericsHomework.csproj
kellanburns: I discovered from a different review that the default for new classes (unless otherwise specified) is to be a class library, so this isn't necessary but wouldn't hurt either, I'm leaving it out though.

GenericsHomework/GenericsHomework.csproj
kellanburns: See response to your other comment, I'm leaving it in because it doesn't hurt.

GenericsHomework/Node.cs
kellanburns: I'm going to keep it just for an added level of assurance that the program won't break in a more unexpected way from me removing the check.

GenericsHomework/Node.cs
kellanburns: I'm going to keep it like this, I think that the generic T here is well understood given the context and the assignment and looks cleaner.

GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,12 @@
...
+    <TargetFramework>net9.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
+    <Nullable>enable</Nullable>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <ImplicitUsings>enable</ImplicitUsings>
```
Joshua-Lester3, line 9: Kind of in line with another comment, I would remove these, since the Directory props file specifies them. Say in the future you change the Directory props file, you'd also have to change this, since these will override the props file.

Joshua-Lester3: also, I think it would be best to specify that it's a library. Even though it defaults to one without a main, it would make it clearer, and, I'm not sure if this would be the case but maybe if you added a Main perhaps it would change? Not fully sure.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+    public Node<T> Next
+    {
+        get => _next;
+        private set => _next = value ?? throw new ArgumentNullException(nameof(value));
+    }
+
+    public Node(T value)
+    {
+        Value = value;
+        _next = this;
```
Joshua-Lester3, line 20: Guideline:
> AVOID accessing the backing field of a property outside the property, even from within the containing class.

Joshua-Lester3 (commented):
## Instructions

- Create a _class library_ project called "GenericsHomework.". ❌ Not specified
  - Define the `Node` class

## Extra Credit

1. Implement a `VennDiagram` structure that contains `n` `Circle`s that only contains homogenous **reference types** of any type. ❌ not implemented

1. Implement `System.Collections.Generic.ICollection<T>` on the `Node` class ❌ not implemented

## Fundamentals

PR #95, opened 2025-12-01
Assignment 9+10

katmilton:
I worked with @Al3xramirez. We implemented the extra credit.


Assignment/PingProcess.cs
```diff
@@ -0,0 +1,313 @@
...
+            (stringBuilder??=new StringBuilder()).AppendLine(line);
+        
+        int exit = RunProcessInternal(StartInfo, updateStdOutput, updateStdOutput, default);
+        return new PingResult( exit, stringBuilder?.ToString());
+    }
+
+    public Task<PingResult> RunTaskAsync(string hostNameOrAddress)
+    {
+        return Task.Run(() =>
+        {
```
chanderlud: It seems like calling the `Run` method in `PingProcess` here would reduce code duplication given that the inside of Task.Run here matches the `Run` method exactly.

Assignment/PingProcess.cs
```diff
@@ -0,0 +1,313 @@
...
+            StringBuilder? stringBuilder = null;
+            void updateStdOutput(string? line) =>
+                (stringBuilder ??= new StringBuilder()).AppendLine(line);
+
+           int exit = RunProcessInternal(StartInfo, updateStdOutput, updateStdOutput, default);
+            return new PingResult(exit, stringBuilder?.ToString());
+        });
+    }
+
+    async public Task<PingResult> RunAsync(
```
chanderlud, line 44: Similarly, the `RunTaskAsync` method could be called from inside this `RunAsync` method to reduce code duplication.

Assignment/PingProcess.cs
```diff
@@ -0,0 +1,313 @@
...
+                    (stringBuilder ??= new StringBuilder()).AppendLine(line);
+                }
+
+                progress.Report(line);
+
+            }
+
+            cancellationToken.ThrowIfCancellationRequested();
+            
+            int process = RunProcessInternal(startInfo, updateStdOutput, default, cancellationToken);
```
chanderlud: I think that it would be appropriate to also pass a delegate to `progressError` here (instead of passing default) so that the `progress` would also receive error messages.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,250 @@
...
+
+    [TestMethod]
+    public void Run_CaptureStdOutput_Success()
+    {
+        PingResult result = Sut.Run("localhost");
+        AssertValidPingOutput(result);
+    }
+
+    [TestMethod]
+    public void RunTaskAsync_Success()
```
chanderlud: This test doesn't follow the naming convention

chanderlud (commented):
1. Implement `PingProcess`' `public Task<PingResult> RunTaskAsync(string hostNameOrAddress)` ✔
   - First implement `public void RunTaskAsync_Success()` test method to test `PingProcess.RunTaskAsync()` using `"localhost"`. ✔
   - Do NOT use async/await in this implementation. ✔
2. Implement `PingProcess`' `async public Task<PingResult> RunAsync(string hostNameOrAddress)` ✔
   - First implement the `public void RunAsync_UsingTaskReturn_Success()` test method to test `PingProcess.RunAsync()` using `"localhost"` **without** using async/await. ✔
   - Also implement the `async public Task RunAsync_UsingTpl_Success()` test method to test `PingProcess.RunAsync()` using `"localhost"` but this time **DO** using async/await. ✔
3. Add support for an optional cancellation token to the `PingProcess.RunAsync()` signature. ✔
   Inside the `PingProcess.RunAsync()` invoke the token's `ThrowIfCancellationRequested()` method so an exception is thrown. ✔
   Test that, when cancelled from the test method, the exception thrown is an `AggregateException` ✔ with a `TaskCanceledException` inner exception ✔ using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping` ✔and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ✔ respectively.
4. Complete/fix **AND test** `async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)` which executes ping for an array of hostNameOrAddresses (which can all be "localhost") **in parallel**, adding synchronization if needed. ✔
   NOTE:
      - The order of the items in the stdOutput is irrelevant and expected to be intermingled.
      - StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔
5. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. Returning a `Task<PingResult>` is also okay.
   NOTE: This method does **NOT** use `Task.Run`.

## Extra Credit

- Test and implement `PingProcess.RunAsync(System.IProgress<T> progess)` so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ✔

## Fundamentals

- Place all shared project items into a `Directory.Build.targets` file.
- Ensure nullable reference types is enabled  ✔
- Set `LangVersion` and the `TargetFramework` to the latest released versions available (preview versions optional)   ✔
- For this assignment, consider using `Assert.AreEqual<T>()` (the generic version)  ✔

Looks like all the assignment requirements are taken care of correctly, I left a few comments on minor things I thought could be improved.

Assignment/PingProcess.cs
```diff
@@ -0,0 +1,313 @@
...
+            string? stdOutput = stringBuilder?.ToString(); 
+            return Task.FromResult (new PingResult(process, stdOutput));
```
thigiang16: Using Task.FromResult inside Task.Run is unnecessary because it creates `Task<Task<PingResult>>`, which is inefficient. You can just do `return new PingResult(process, stdOutput);`

Assignment/PingProcess.cs
```diff
@@ -0,0 +1,313 @@
...
+    IEnumerable<string> hostNameOrAddresses,
+    CancellationToken cancellationToken = default)
+    {
+        ArgumentNullException.ThrowIfNull(hostNameOrAddresses);
+
+        StringBuilder stringBuilder = new();
+        object syncLock = new();
+
+        Task<int>[] tasks = hostNameOrAddresses
+            .Select(host => Task.Run(() =>
```
thigiang16: I think using Task.Run inside .Select adds complexity. You can call your async method inside the Select and then use Task.WhenAll to run all tasks in parallel. This keeps it simpler without extra task wrapping

Assignment/PingProcess.cs
```diff
@@ -0,0 +1,313 @@
...
+
+        void updateStdOutput(string? line) =>
+            (stringBuilder ??= new StringBuilder()).AppendLine(line);
+
+        int exit = await Task.Run(() =>
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            return RunProcessInternal(StartInfo, updateStdOutput, default, cancellationToken);
+
+        }, cancellationToken);
```
thigiang16: I think you don't need to throw inside Task.Run. Using `WaitAsync(cancellationToke)` lets the task handle cancellation properly

thigiang16 (commented):
Instructions 
1.Implement PingProcess' public Task<PingResult> RunTaskAsync(string hostNameOrAddress) ✔
First implement public void RunTaskAsync_Success() test method to test PingProcess.RunTaskAsync() using "localhost". ✔ 
Do NOT use async/await in this implementation. ✔ 
2. Implement PingProcess' async public Task<PingResult> RunAsync(string hostNameOrAddress) ✔ 
First implement the public void RunAsync_UsingTaskReturn_Success() test method to test PingProcess.RunAsync() using "localhost" without using async/await. ✔ 
Also implement the async public Task RunAsync_UsingTpl_Success() test method to test PingProcess.RunAsync() using "localhost" but this time DO using async/await.✔ 
3. Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔ Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown.✔ 
Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔ with a TaskCanceledException inner exception ✔ using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ✔and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively. 
4. Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE: 
The order of the items in the stdOutput is irrelevant and expected to be intermingled. StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔ 
5. Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run. ✔ 
Extra Credit 
Test and implement PingProcess.RunAsync(System.IProgress<T> progess) so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ✔ Fundamentals 
Place all shared project items into a Directory.Build.targets file. Ensure nullable reference types is enabled ✔ 
Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔ and enabled .NET analyzers for both projects ✔ 
For this assignment, consider using Assert.AreEqual<T>() (the generic version) ✔ 
Choose simplicity over complexity ❌

The implementation looks good. A few areas could be simplified to favor simplicity over complexity, like avoiding unnecessary Task.Run or reducing duplicate code. Consider reusing existing methods to make it cleaner. You might also add more tests to improve coverage.

Hellothereyoko:
Instructions
1.Implement PingProcess' public Task RunTaskAsync(string hostNameOrAddress) ✔
First implement public void RunTaskAsync_Success() test method to test PingProcess.RunTaskAsync() using "localhost". ✔
Do NOT use async/await in this implementation. ✔
2. Implement PingProcess' async public Task RunAsync(string hostNameOrAddress) ✔
First implement the public void RunAsync_UsingTaskReturn_Success() test method to test PingProcess.RunAsync() using "localhost" without using async/await. ✔
Also implement the async public Task RunAsync_UsingTpl_Success() test method to test PingProcess.RunAsync() using "localhost" but this time DO using async/await.✔
3. Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔ Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown.✔
Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔ with a TaskCanceledException inner exception ✔ using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ✔and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively.
4. Complete/fix AND test async public Task RunAsync(IEnumerable hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE:
The order of the items in the stdOutput is irrelevant and expected to be intermingled. StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔
5. Implement AND test public Task RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task is also okay. NOTE: This method does NOT use Task.Run. ✔
Extra Credit
Test and implement PingProcess.RunAsync(System.IProgress progess) so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ✔ Fundamentals
Place all shared project items into a Directory.Build.targets file. Ensure nullable reference types is enabled ✔
Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔ and enabled .NET analyzers for both projects ✔
For this assignment, consider using Assert.AreEqual() (the generic version) ✔
Choose simplicity over complexity ❌

The implementation looks good but try to save task.runs bc they eat memory

katmilton:
> Instructions 1.Implement PingProcess' public Task RunTaskAsync(string hostNameOrAddress) ✔ First implement public void RunTaskAsync_Success() test method to test PingProcess.RunTaskAsync() using "localhost". ✔ Do NOT use async/await in this implementation. ✔ 2. Implement PingProcess' async public Task RunAsync(string hostNameOrAddress) ✔ First implement the public void RunAsync_UsingTaskReturn_Success() test method to test PingProcess.RunAsync() using "localhost" without using async/await. ✔ Also implement the async public Task RunAsync_UsingTpl_Success() test method to test PingProcess.RunAsync() using "localhost" but this time DO using async/await.✔ 3. Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔ Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown.✔ Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔ with a TaskCanceledException inner exception ✔ using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ✔and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively. 4. Complete/fix AND test async public Task RunAsync(IEnumerable hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE: The order of the items in the stdOutput is irrelevant and expected to be intermingled. StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔ 5. Implement AND test public Task RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task is also okay. NOTE: This method does NOT use Task.Run. ✔ Extra Credit Test and implement PingProcess.RunAsync(System.IProgress progess) so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ✔ Fundamentals Place all shared project properties into a Directory.Build.Props file. Place all shared project items into a Directory.Build.targets file. Ensure nullable reference types is enabled ✔ Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ✔ Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔ and enabled .NET analyzers for both projects ✔ For this assignment, consider using Assert.AreEqual() (the generic version) ✔ All of the above should be unit tested ✔ Choose simplicity over complexity ❌
> 
> The implementation looks good but try to save task.runs bc they eat memory

Thanks for the suggestion! I did experiment with removing some of the Task.Run usages, but the refactors ended up breaking the CI tests as they changed the expected execution behavior. This would be a great candidate for optimization and refactoring in the future though.

Assignment/PingProcess.cs
katmilton: Good call — removing duplication would normally improve maintainability. In this case I tried refactoring to reuse methods, but it caused CI failures because the tests rely on the original ordering and behavior inside the async wrappers. To keep all tests passing, I restored the original structure. Thank you for the feedback.

Assignment/PingProcess.cs
katmilton: Good call — removing duplication would normally improve maintainability. In this case I tried refactoring to reuse methods, but it caused CI failures because the tests rely on the original ordering and behavior inside the async wrappers. To keep all tests passing, I restored the original structure. Thank you for the feedback.

Assignment.Tests/PingProcessTests.cs
katmilton: Thanks! I've updated the test name to better follow naming guidelines.

Assignment/PingProcess.cs
katmilton: Totally agree this would improve visibility during actual ping execution. For the assignment, the existing tests only validate stdout progress, so altering stderr handling changed the behavior they expect. I reverted that change to stay aligned with the assignment’s requirements. Thanks for the feedback!

Assignment/PingProcess.cs
katmilton: Great point. That definitely improves clarity. I removed that part of the refactor since it wasn’t needed to pass the tests, but it would definitely be a good idea to implement in future iterations. Thank you for the feedback.

Assignment/PingProcess.cs
katmilton: Thanks for the feedback. I tested this change, but it caused differences in exception flow and timing that made cancellation-related tests fail. I’ll keep that approach in mind for future refactors in a non-strict test environment.

Assignment/PingProcess.cs
katmilton: Thanks for the suggestion. I agree that removing the extra Task.Run inside .Select would normally simplify the implementation and avoid unnecessary task wrapping. I did try refactoring this to directly call the async overload and rely on Task.WhenAll, but it ended up changing the execution behavior enough that CI tests started failing, especially around how timing and output aggregation are validated. For now, the change is reverted but this would be a great thing to improve upon in future iterations.

JosephPotapenko:
Implement PingProcess' public Task RunTaskAsync(string hostNameOrAddress) ✔
First implement public void RunTaskAsync_Success() test method to test PingProcess.RunTaskAsync() using "localhost". ✔
Do NOT use async/await in this implementation. ✔
Implement PingProcess' async public Task RunAsync(string hostNameOrAddress) ✔
First implement the public void RunAsync_UsingTaskReturn_Success() test method to test PingProcess.RunAsync() using "localhost" without using async/await. ✔
Also implement the async public Task RunAsync_UsingTpl_Success() test method to test PingProcess.RunAsync() using "localhost" but this time DO using async/await. ✔
Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔
Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ✔
Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔
with a TaskCanceledException inner exception ✔
using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ✔and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively.
Complete/fix AND test async public Task RunAsync(IEnumerable hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE:
The order of the items in the stdOutput is irrelevant and expected to be intermingled.
StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔
Implement AND test public Task RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task is also okay. NOTE: This method does NOT use Task.Run.

Extra Credit
Test and implement PingProcess.RunAsync(System.IProgress progess) so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ✔

Fundamentals
Place all shared project items into a Directory.Build.targets file.
Ensure nullable reference types is enabled ✔
For this assignment, consider using Assert.AreEqual() (the generic version) ✔

Good work. There's not much to mention than what's already been brought up, so I'll leave it at that!

Assignment.Tests/StringExtensions.cs
```diff
@@ -1,35 +1,36 @@
...
-    // about escapeCharacter is required.  
-    public static bool IsLike(this string text, string pattern) =>
-        new WildcardPattern(pattern).IsMatch(text);
-
-    /// <summary>
-    /// Implement's VB's Like operator logic.
-    /// </summary>
-    public static bool IsLike(this string text, string pattern, char escapeCharacter) =>
-        new WildcardPattern(pattern, escapeCharacter).IsMatch(text);
+﻿#pragma warning disable
```
Joshua-Lester3, line 1: Maybe Ben said to do this, so let me know if I missed that. If not, this either needs explanation, or to be removed.

Assignment.Tests/WildCardPattern.cs
```diff
@@ -0,0 +1,1240 @@
+﻿#pragma warning disable
```
Joshua-Lester3, line 1: Same here

Assignment/ClonePing.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Assignment;
+
+public class ClonePing : PingProcess
```
Joshua-Lester3, line 10: I might prefer either keeping this in the Test project, or using a mocking library in your tests.

Assignment/PingProcess.cs
```diff
@@ -0,0 +1,313 @@
...
+
+    public PingResult Run(string hostNameOrAddress)
+    {
+        StartInfo.Arguments = BuildPingArguments(hostNameOrAddress);
+        StringBuilder? stringBuilder = null;
+
+        void updateStdOutput(string? line) =>
+            (stringBuilder??=new StringBuilder()).AppendLine(line);
+        
+        int exit = RunProcessInternal(StartInfo, updateStdOutput, updateStdOutput, default);
```
Joshua-Lester3, line 26: Is there a reason `default` was switched to `updateStdOutput` for the third param?

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,241 @@
...
+        }
+        catch (AggregateException ex)
+        {
+
+            Assert.IsInstanceOfType<AggregateException>(ex);
+            return;
+
+        }
+
+        Assert.Fail("Expected Aggregate Exception, but none was thrown");
```
Joshua-Lester3, line 112: I'd probably do this using Assert.Throws: https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.assert.throwsexception?view=mstest-net-3.11

Joshua-Lester3 (commented):
## Instructions

1. Implement `PingProcess`' `public Task<PingResult> RunTaskAsync(string hostNameOrAddress)` ✔
   - First implement `public void RunTaskAsync_Success()` test method to test `PingProcess.RunTaskAsync()` using `"localhost"`. ✔
   - Do NOT use async/await in this implementation. ✔
2. Implement `PingProcess`' `async public Task<PingResult> RunAsync(string hostNameOrAddress)` ✔
   - First implement the `public void RunAsync_UsingTaskReturn_Success()` test method to test `PingProcess.RunAsync()` using `"localhost"` **without** using async/await. ✔
   - Also implement the `async public Task RunAsync_UsingTpl_Success()` test method to test `PingProcess.RunAsync()` using `"localhost"` but this time **DO** using async/await. ✔
3. Add support for an optional cancellation token to the `PingProcess.RunAsync()` signature. ✔
   Inside the `PingProcess.RunAsync()` invoke the token's `ThrowIfCancellationRequested()` method so an exception is thrown. ✔
   Test that, when cancelled from the test method, the exception thrown is an `AggregateException` ✔ with a `TaskCanceledException` inner exception ✔ using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping` ✔and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ✔ respectively.
4. Complete/fix **AND test** `async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)` which executes ping for an array of hostNameOrAddresses (which can all be "localhost") **in parallel**, adding synchronization if needed. ✔
   NOTE:
      - The order of the items in the stdOutput is irrelevant and expected to be intermingled.
      - StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔
5. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. Returning a `Task<PingResult>` is also okay.
   NOTE: This method does **NOT** use `Task.Run`.

## Extra Credit

- Test and implement `PingProcess.RunAsync(System.IProgress<T> progess)` so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ✔

## Fundamentals

- Place all shared project items into a `Directory.Build.targets` file.
- Ensure nullable reference types is enabled  ✔
- Set `LangVersion` and the `TargetFramework` to the latest released versions available (preview versions optional)   ✔
- For this assignment, consider using `Assert.AreEqual<T>()` (the generic version)  ✔

Great job!

Assignment.Tests/PingProcessTests.cs
Joshua-Lester3: Good catch - this was actually in the requirements though haha. Obviously no points off, because you're actually fixing it - feel free to make a PR to fix it though :) Pretty sure Ben would happily accept that.

